PR #7, opened 2024-05-08
adding reference to frontend tsconfig in client config.

jchoi330:


packages/client/tsconfig.json
```diff
@@ -1,4 +1,19 @@
...
+    "strict": true,
+    "skipLibCheck": true,
+    "moduleResolution": "node",
+    "baseUrl": ".",
+    "esModuleInterop": true,
+    "forceConsistentCasingInFileNames": true,
+
+    "noEmit": true
+  },
+  "references": [
```
jchoi330, line 12: https://www.typescriptlang.org/docs/handbook/project-references.html

jchoi330: Need to run this with the build flag: `tsc -b`

wanderer: this disable emission of files if there is an error. I guess this is ok to enable. Intresting that this solve the problem.

PR #8, opened 2024-05-08
enforcing ts checks in client dir

jchoi330:


packages/client/test/relayClient.test.ts
```diff
@@ -52,7 +52,6 @@ beforeEach(async () => {
   crypto.getRandomValues(keyCard);
   relayClient = new RelayClient({
     relayEndpoint,
-    useTLS: false,
```
jchoi330, line 55: don't need to specify this, this is determined from relay endpoint in the client

packages/client/test/mockClient.ts
```diff
@@ -32,8 +32,8 @@ export type VectorItems = {
 export type TestVectors = {
   signatures: {
     chain_id: number;
-    contract_address: `0x${string}`;
-    signer_address: `0x${string}`;
+    contract_address: string;
```
jchoi330, line 35: testVectors are generated with the 0x hex prefix.

packages/client/tsconfig.json
```diff
@@ -1,4 +1,4 @@
 {
-  "extends": ["../../tsconfig.json"],
+  "extends": ["../../tsconfig.json", "../frontend/tsconfig.json"],
```
jchoi330, line 2: this resolved the alias issue.

PR #15, opened 2024-05-13
moving #authenticate to connect and setting auth prop

jchoi330:


packages/client/lib/index.ts
```diff
@@ -123,6 +124,7 @@ export class RelayClient extends EventEmitter {
       verifyingContract: abi.addresses.StoreReg as Address,
     };
     this.firstEvent = null;
+    this.authenticated = true;
```
jchoi330, line 127: Only set this to false if websocket is disconnected.

packages/client/lib/index.ts
```diff
@@ -284,7 +299,13 @@ export class RelayClient extends EventEmitter {
         this.#decodeMessage.bind(this),
       );
     }
-
+    if (
+      this.connection.readyState === WebSocket.OPEN &&
```
wanderer, line 303: we should use the "open" event and then authenticate, instead of checking the state each time. It think. So we should remove this.

PR #25, opened 2024-05-28
walletconnect QR scan

jchoi330:


packages/frontend/src/app/components/transactions/TransactionStatus.tsx
```diff
@@ -123,6 +124,10 @@ const TransactionStatus = ({
           totalDollar={totalDollar}
           purchaseAddress={purchaseAddress}
         />
+        {/* <WalletConnectQR
```
jchoi330, line 127: commenting this out for now till we get the design for the payment flow

PR #27, opened 2024-05-30
using recoverTypedDataAddress

jchoi330:


packages/frontend/src/app/components/login/ConnectWallet.tsx
```diff
@@ -51,6 +51,10 @@ const ConnectWallet = ({ close }: { close: () => void }) => {
     _wallet && setWallet(_wallet);
   }, [walletStatus]);
 
+  useEffect(() => {
+    enrollKeycard.current = false;
+  }, [clientWallet]);
+
```
jchoi330, line 57: small bug fix: If user changes the connected wallet, we should re-emit the keycard enroll event

packages/frontend/src/utils/buildState.ts
```diff
@@ -206,6 +210,30 @@ export const buildState = (
...
+          cartId,
+          signature,
+        },
+      });
+    } else if (e.newKeyCard) {
+      const userWalletAddr = bytesToHex(e.newKeyCard.userWalletAddr);
+      const cardPublicKey = bytesToHex(e.newKeyCard.cardPublicKey);
+      if (
+        walletAddress &&
+        walletAddress.toLowerCase() == userWalletAddr.toLowerCase()
```
jchoi330, line 228: saving all the key card public keys associated with the current wallet address

packages/frontend/src/reducers/cartReducers.ts
```diff
@@ -123,7 +135,11 @@ function cartStateReducer(
         ...state,
         items: {},
       };
-
+    case SET_CART_SIG:
```
jchoi330, line 138: saving the create cart signature

packages/frontend/src/context/StoreContext.tsx
```diff
@@ -202,13 +213,40 @@ export const StoreContextProvider = (
...
+            setPubKeys,
+            walletAddress,
           );
         }
       }
     } catch (err) {
       console.error("error receiving events", err);
     }
   };
+  const verify = async (
```
jchoi330, line 225: calling verifySignedTypeData for every cartId (that have not been invalidated/failed) with every key card address associated with current wallet.

packages/client/lib/protobuf/compiled.d.ts
```diff
@@ -1320,10 +1320,10 @@ export namespace market {
       eventId?: Uint8Array | null;
 
       /** NewKeyCard userWalletAddr */
-      userWalletAddr?: Uint8Array | null;
+      userWalletAddr: Uint8Array;
```
cryptix, line 1323: hmm.. can you point me to why this is necessary? next time the schema changes this needs to be re-done.. so we either need a .patch file to re-apply or script this somehow into the `pb-gen` build step.

packages/frontend/src/context/StoreContext.tsx
cryptix: > every key card address associated with current wallet

Can you expand on this? The event might be signed by some other user, too.

Basically you get the keyCard/public key that signed the event via `ecrecover()` (in python it's called `recover_message`) and then you just need to check if the key is in the set of KeyCards + the assigned relays from the contract. 

```py
    def _valid_event_signing_addresses(self):
        all = []
        # retreive all relays nfts
        if self.storeReg.functions.getRelayCount(self.store_token_id).call() > 0:
            all_relay_token_ids = self.storeReg.functions.getAllRelays(self.store_token_id).call()
            for token_id in all_relay_token_ids:
                # retreive the owner => it's address
                relay_address = self.relayReg.functions.ownerOf(token_id).call()
                all.append(relay_address.lower())
        # turn key cards into addresses
        key_card_addresses = [public_key_to_address(pk).lower() for pk in list(self.all_key_cards.keys())]
        all.extend(key_card_addresses)
        return all

    def _verify_event(self, evt: schema_pb2.Event):
        if len(evt.signature) != 65:
            raise Exception("Invalid signature length: {}".format(len(evt.signature)))
        encoded_data = self.event_hasher.hash_event(evt)
        pub_key = self.w3.eth.account.recover_message(encoded_data, signature=evt.signature)
        print(f"event:{get_id_from_event(evt)} recovered pub_key: {pub_key}")
        their_addr = Web3.to_checksum_address(pub_key).lower()
        valid_addrs = self._valid_event_signing_addresses()
        if their_addr not in valid_addrs:
            print(f"valid addresses: {valid_addrs}")
            raise Exception("Event signed by unknown address: {}".format(their_addr))
```

`self.all_key_cards` should be very similar to the reducer you built.

Main thing is, you shouldn't have to iterate over all the possible signing addresses again and again.

cryptix (changes requested):
Good start!

Maybe martin can help with some TS sparkles to generalize over the event types.

In a sense, since we can sign already it shoudln't be much more work to validate once you keep track of all accepted KeyCards.

in pseudo-code sign is basically:
```
h = hash(event)
sig = sign(h, keyCard.private)
````
verify is:

```
h = hash(event)
keyCardPublic = ecrecover(h, sig)
all.contains(keyCardPublic)
```

File: packages/client/lib/protobuf/compiled.d.ts (hunk 5 above)
Comment (jchoi330) [reply packages/client/lib/protobuf/compiled.d.ts:1323]: right... this is to bypass the typescript error in [buildState](https://github.com/masslbs/Tennessine/pull/27/files#diff-e0ca6374c72417e902d431ff5105811ac53e22b3b601a105ab94f54c390ddf17R224).

it doesn't like that we are trying to call bytesToHex on a value thats possibly undefined

Comment (cryptix) [reply packages/client/lib/protobuf/compiled.d.ts:1323]: Hmm.. can you point me to the line where this is a problem? I think you should be able to work past that by using `event.field!` when passing the value but I'm not sure.

Comment (jchoi330) [reply packages/client/lib/protobuf/compiled.d.ts:1323]: https://github.com/masslbs/Tennessine/pull/27/files#diff-e0ca6374c72417e902d431ff5105811ac53e22b3b601a105ab94f54c390ddf17R224

okay yes i will do the bang instead

File: packages/frontend/src/context/StoreContext.tsx [hunk 6]
```diff
@@ -202,13 +216,40 @@ export const StoreContextProvider = (
...
+      const _cart = _cartItems.get(_cartId) as CartState;
+      if (_cart && _cart.status !== IStatus.Failed) {
+        const sig = _cart.signature as `0x${string}`;
+        const retrievedAdd = await relayClient!.recoverSignedAddress(
+          _cartId,
+          sig,
+        );
+        if (addresses.includes(retrievedAdd.toLowerCase())) {
+          console.log("inside inclue", _cartId);
+          setCartId(_cartId);
```
wanderer, line 248: does this assume the  clerk has only one cart? or if setCartId gets called multiple times do the carts get saved some where else?

jchoi330: Clerk can only have one active cart so setCartId will can get called multiples times but the most recent will be set

wanderer (approved):

PR #28, opened 2024-05-30
emit keycard enroll after wallet connection

jchoi330:


packages/frontend/src/app/components/login/ConnectWallet.tsx
```diff
@@ -51,6 +51,10 @@ const ConnectWallet = ({ close }: { close: () => void }) => {
     _wallet && setWallet(_wallet);
   }, [walletStatus]);
 
+  useEffect(() => {
+    enrollKeycard.current = false;
+  }, [clientWallet]);
+
```
jchoi330, line 57: bug fix to enroll KC if user swaps their wallet connection

packages/frontend/src/context/StoreContext.tsx
```diff
@@ -67,7 +67,11 @@ export const StoreContextProvider = (
   useEffect(() => {
     (async () => {
       const { Level } = await import("level");
-      const db = new Level("./db", { valueEncoding: "json" });
+      const storeId =
```
jchoi330, line 70: set level db name as store id.

wanderer (approved):

PR #48, opened 2024-06-19
getting and verifying storeids

jchoi330:


packages/frontend/src/context/MyContext.tsx
```diff
@@ -102,6 +113,45 @@ export const MyContextProvider = (
     address: address as `0x${string}`,
   });
 
+  const getStores = async () => {
+    const stores = new Map();
+
+    const logs = await publicClient.getLogs({
+      address: abi.addresses.ShopReg as Address,
```
wanderer, line 120: we want to get only the event  `Transfer` so we need to add something like  `event: abi.Transfer`  That way we only get the transfer events. Transfer emits the following     `event Transfer(address indexed _from, address indexed _to, uint256 indexed _tokenId);`  so you can build the map of merchants to shops without anyother queries

packages/frontend/src/context/MyContext.tsx
```diff
@@ -102,6 +119,47 @@ export const MyContextProvider = (
     address: address as `0x${string}`,
   });
 
+  const getStores = async () => {
+    const stores = new Map();
+
+    const logs = await publicClient.getLogs({
```
wanderer, line 125: So the default behavior of getLogs is to only get the Logs from the last block 
https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getlogs
```
fromBlock: QUANTITY|TAG - (optional, default: "latest") Integer block number, or "latest" for the last mined block, "safe" for the latest safe block, "finalized" for the latest finalized block, or "pending", "earliest" for not yet mined transactions.

toBlock: QUANTITY|TAG - (optional, default: "latest") Integer block number, or "latest" for the last mined block, "safe" for the latest safe block, "finalized" for the latest finalized block, or "pending", "earliest" for not yet mined transactions.
```
It look like it is assumed that this will return all the logs.

packages/frontend/src/context/MyContext.tsx
```diff
@@ -102,6 +119,47 @@ export const MyContextProvider = (
...
+      address: abi.addresses.ShopReg as Address,
+      event: parseAbiItem(
+        "event Transfer(address indexed from, address indexed to, uint256 value)",
+      ),
+    });
+    const _stores = logs[0].topics;
+    await Promise.all(
+      _stores.map(async (id) => {
+        try {
+          const ownerAdd = (await publicClient.readContract({
```
wanderer, line 135: The logs already have have owner `to` in them (`address indexed to`) you should be able to extract this infromation from the logs directly without having to query the contract again.

PR #57, opened 2024-06-25
metadata blob update

jchoi330:


packages/frontend/src/app/create-store/page.tsx
```diff
@@ -95,16 +95,49 @@ const StoreCreation = () => {
...
+            description: {
+              type: "string",
+              description,
+            },
+            image: {
+              type: "string",
+              description: path.url,
+            },
+          },
+        };
```
cryptix, line 129: The schema/format here needs a little fixing up.

https://eips.ethereum.org/EIPS/eip-721 lists it as "JSON schema" which defines what the fields mean (description) and what type they are.

The _actual_ value would just be a flat object ala

```
{
  "name": "my store",
  "description": "cat nip and other fun things",
  "image": "https://you.get/the/idea.png"
}
```

here is a more elaborate example going from value to schema: https://json-schema.org/learn/getting-started-step-by-step

packages/client/src/blockchainClient.ts
```diff
@@ -36,6 +36,15 @@ export class BlockchainClient {
     });
   }
 
+  setShopTokenId(wallet: WalletClientWithAccount, uri: string) {
```
wanderer, line 39: should be setShopsURI maybe

wanderer (approved):

PR #67, opened 2024-07-01
Copy jul 1

jchoi330:
1. setting base currency address
2. fixed add/editing product functions that were broken because of hex vs. bytes 
3. adding payee address

### DIFF
--- a/packages/frontend/src/context/StoreContext.tsx
+++ b/packages/frontend/src/context/StoreContext.tsx
@@ -128,7 +128,6 @@ export const StoreContextProvider = (
         )) as OrdersMap;
 
         const orderIdLocal = await getItem("orderId", db);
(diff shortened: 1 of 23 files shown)


packages/frontend/src/context/StoreContext.tsx
```diff
@@ -118,46 +128,60 @@ export const StoreContextProvider = (
...
 
         if (productsLocal?.size) {
           setProducts({
             type: SET_PRODUCTS,
             payload: {
-              itemId: productsLocal.keys().next().value,
               allProducts: productsLocal,
             },
           });
+        } else {
```
jchoi330, line 141: resetting all store state when a new store is created

wanderer (approved):

PR #69, opened 2024-07-02
comment out block range for sepolia

jchoi330:


packages/frontend/src/context/StoreContext.tsx
```diff
@@ -429,7 +436,10 @@ export const StoreContextProvider = (
 
   const addProductToTag = async (tagId: TagId, itemId: ItemId) => {
     try {
-      await relayClient!.updateTag({ tagId, addItemId: itemId });
+      await relayClient!.updateTag({
+        tagId: hexToBytes(tagId),
+        removeItemId: hexToBytes(itemId),
```
cryptix, line 441: i think this needs to be `addItemId`

cryptix (changes requested):
i think there is bug/wrong field name in one of the updateTag() calls but rest looks good!

packages/frontend/src/context/StoreContext.tsx
jchoi330: oh good catch

cryptix (approved):
👏🏻

PR #79, opened 2024-07-05
walletconnect from address change

jchoi330:


packages/frontend/src/app/components/transactions/WalletConnectQR.tsx
```diff
@@ -69,7 +69,10 @@ function WalletConnectQR({
         const sessionNamespace = await approval();
         if (sessionNamespace) {
           setSession(sessionNamespace);
-          setAccount(sessionNamespace.namespaces.eip155.accounts[0].slice(9));
+          const accountAddress = sessionNamespace.namespaces.eip155.accounts[0]
+            .slice(9)
```
cryptix, line 73:
```suggestion
            .slice(9) // split of 'ethereum:'  prefix
```

seems a bit to magical without a comment

PR #80, opened 2024-07-18
[wip] StateManager

jchoi330:


packages/stateManager/index.ts
```diff
@@ -36,11 +44,10 @@ abstract class PublicObjectManager<
     super();
   }
   abstract _processEvent(event: schema.ShopEvents): Promise<boolean>;
-  abstract create(obj: T): Promise<[any, any]>;
-  abstract update(obj: T): Promise<[any, any]>;
-  get(id: string): Promise<T> {
-    return this.db.get(id);
-  }
+  abstract createItem(obj: T): Promise<void>;
```
wanderer, line 47:
```suggestion
  abstract create(obj: T): Promise<void>;
```

packages/stateManager/index.ts
```diff
@@ -36,11 +44,10 @@ abstract class PublicObjectManager<
...
+  abstract updateItem(obj: T): Promise<void>;
```
wanderer, line 48:
```suggestion
  abstract update(obj: T): Promise<void>;
```

packages/stateManager/index.ts
```diff
@@ -36,11 +44,10 @@ abstract class PublicObjectManager<
...
+  abstract getAllListings(): Promise<Map<string, Item>>;
+  abstract getListing(id: string): Promise<any | null>;
```
wanderer, line 50:
```suggestion
  abstract getAll(): Promise<Map<string, Item>>;
  abstract get(id: string): Promise<any | null>;
```

packages/stateManager/index.ts
```diff
@@ -50,30 +57,117 @@ class ItemManager extends PublicObjectManager<Item> {
...
+    return false;
   }
-  create(item: Item) {
-    return Promise.all([
-      this.db.put(item.id, item),
-      this.client.createItem(item),
-    ]);
+  async createItem(item: Item) {
+    await this.client.createItem(item);
+    await this.db.put(`products-${item.id}`, item);
```
wanderer, line 146:
```suggestion
    await this.db.put(item.id, item);
```

packages/stateManager/index.ts
```diff
@@ -36,67 +77,293 @@ abstract class PublicObjectManager<
...
+          const diff = cs.diffs ? cs.diffs[i] : 0;
+          item.quantity = item.quantity + diff;
+          await this.db.put(itemId, item);
+        });
+        // if the changeStock event is emitted from a payment(AKA has txHash), it has to also go through OrderStore
+        if (cs.txHash) {
+          return false;
+        } else return true;
+      }
+    } else if (event.updateTag) {
```
jchoi330, line 130: we need these tag events in the product sub level since we are adding and removing tags from a product

packages/stateManager/index.ts
```diff
@@ -36,67 +101,345 @@ abstract class PublicObjectManager<
     super();
   }
   abstract _processEvent(event: schema.ShopEvents): Promise<boolean>;
-  abstract create(obj: T): Promise<[any, any]>;
-  abstract update(obj: T): Promise<[any, any]>;
-  get(id: string): Promise<T> {
-    return this.db.get(id);
-  }
+  abstract get(key: string): Promise<T>;
+  abstract getAll(): T[];
```
wanderer, line 105: just return iterator

packages/stateManager/index.ts
```diff
@@ -36,67 +101,345 @@ abstract class PublicObjectManager<
...
   async _processEvent(event: schema.ShopEvents): Promise<boolean> {
-    if (typeof event === schema.ShopEvents.UpdateItem) {
-      // load the item from the store
-      const item = await this.db.get(event.eventId);
-      Object.assign(item, event);
-      await this.db.put(event.eventId, item);
-      this.emit("update", item);
+    if (event.createItem) {
+      const ci = event.createItem;
+      const metadata = JSON.parse(new TextDecoder().decode(ci.metadata));
```
wanderer, line 117: UintArrayToJSON

wanderer (commented):
It would be good to have a comment for each public function that describes what it does.

packages/stateManager/index.ts
```diff
@@ -36,67 +101,345 @@ abstract class PublicObjectManager<
...
+      }
+      if (ui.price) {
+        item.price = ui.price;
+      }
+      await this.db.put(id, item);
+      return true;
+    } else if (event.changeStock) {
+      const cs = event.changeStock;
+      if (cs.itemIds) {
+        cs.itemIds.map(async (id: Uint8Array, i: number) => {
```
wanderer, line 142: forEach not map

packages/stateManager/index.ts
```diff
@@ -36,67 +101,345 @@ abstract class PublicObjectManager<
...
+      const metadata = JSON.parse(new TextDecoder().decode(ci.metadata));
+      const id = bytesToHex(ci.eventId);
+      const item = {
+        price: ci.price,
+        metadata,
+        tags: [],
+        quantity: 0,
+      };
+      await this.db.put(id, item);
+      return true;
```
wanderer, line 126:
```suggestion

```

packages/stateManager/index.ts
```diff
@@ -36,67 +101,345 @@ abstract class PublicObjectManager<
...
+      const id = bytesToHex(ui.itemId);
+      const item = await this.db.get(id);
+      if (ui.metadata) {
+        item.metadata = JSON.parse(new TextDecoder().decode(ui.metadata));
+      }
+      if (ui.price) {
+        item.price = ui.price;
+      }
+      await this.db.put(id, item);
+      return true;
```
wanderer, line 138:
```suggestion
```

packages/stateManager/index.ts
```diff
@@ -36,67 +101,345 @@ abstract class PublicObjectManager<
...
+    const id = bytesToHex(
+      await this.client.createItem({
+        price: item.price,
+        metadata: new TextEncoder().encode(JSON.stringify(item.metadata)),
+      }),
+    );
+    await this.db.put(id, item);
+  }
+
+  async update(
```
wanderer, line 185: `update` should handle adding and removing tags

packages/stateManager/index.ts
```diff
@@ -36,67 +101,345 @@ abstract class PublicObjectManager<
...
+    itemId: `0x${string}`,
+    price?: string,
+    metadata?: { name: string; description: string; image: string },
+    quantity?: number,
+  ) {
+    const item = await this.db.get(itemId);
```
wanderer, line 191: do this in parallel with network operation so that when the relay has confirmed the update we will have the data from the db.

packages/stateManager/index.ts
```diff
@@ -36,67 +101,345 @@ abstract class PublicObjectManager<
...
+    if (price) {
+      item.price = price;
+      await this.client.updateItem({ itemId: hexToBytes(itemId), price });
```
wanderer, line 194: we should only call `updateItem` once per call to `this.update`

packages/stateManager/index.ts
```diff
@@ -36,67 +101,345 @@ abstract class PublicObjectManager<
...
+    if (metadata) {
+      item.metadata = metadata;
+      await this.client.updateItem({
+        itemId: hexToBytes(itemId),
+        metadata: new TextEncoder().encode(JSON.stringify(metadata)),
+      });
+    }
+    if (quantity) {
+      const diff = quantity - item.quantity;
+      this.client.changeStock({ itemIds: [hexToBytes(itemId)], diffs: [diff] });
```
wanderer, line 205: we need to wait for this promise to resolve. Ideally this would be done concurrently with `client.updateItem`

packages/stateManager/index.ts
```diff
@@ -36,67 +101,345 @@ abstract class PublicObjectManager<
...
+      item.quantity = quantity;
+    }
+
+    await this.db.put(itemId, item);
+  }
+
+  getAll() {
```
wanderer, line 212: should just proxy the iterator.

packages/stateManager/index.ts
```diff
@@ -36,67 +101,345 @@ abstract class PublicObjectManager<
...
+    return this.db.iterator().all();
+  }
+
+  get(key: `0x${string}`) {
+    return this.db.get(key);
+  }
+}
+
+class ShopDetailsManager extends PublicObjectManager<ShopFields> {
```
wanderer, line 221: Manifest

packages/stateManager/index.ts
```diff
@@ -36,67 +101,345 @@ abstract class PublicObjectManager<
...
+  constructor(db: Store<ShopFields>, client: RelayClient) {
```
wanderer, line 222:
```suggestion
  constructor(store: Store<ShopFields>, client: RelayClient) {
```

wanderer (changes requested):
It would be good to have a comment for each public function that describes what it does.

PR #82, opened 2024-07-22
adding state manager to workspace

jchoi330:
adds Uint8ArrayToParsedJSON to utils

### DIFF
--- a/packages/stateManager/index.ts
+++ b/packages/stateManager/index.ts
@@ -2,7 +2,7 @@ import { EventEmitter } from "events";
 import { RelayClient } from "@massmarket/client";
 import schema from "@massmarket/schema";
 import { bytesToHex, hexToBytes } from "viem";
-import { Uint8ArrayToParsedJSON } from "@massmarket/utils";
+import { bufferToJSON } from "@massmarket/utils";
 interface Item {
   id: `0x${string}`;
   price: string;
@@ -111,17 +111,17 @@ class ListingManager extends PublicObjectManager<Item> {
       const item = {
         id,
         price: ci.price,
-        metadata: Uint8ArrayToParsedJSON(ci.metadata),
+        metadata: bufferToJSON(ci.metadata),
         tags: [],
         quantity: 0,
       };
-      await this.store.put(id, item);
+      return this.store.put(id, item);
     } else if (event.updateItem) {
       const ui = event.updateItem;
       const id = bytesToHex(ui.itemId);
       const item = await this.store.get(id);
       if (ui.metadata) {
-        item.metadata = Uint8ArrayToParsedJSON(ui.metadata);
+        item.metadata = bufferToJSON(ui.metadata);
       }
       if (ui.price) {
         item.price = ui.price;
@@ -135,7 +135,7 @@ class ListingManager extends PublicObjectManager<Item> {
           const item = await this.store.get(itemId);
           const diff = cs.diffs ? cs.diffs[i] : 0;
           item.quantity = item.quantity + diff;
-          await this.store.put(itemId, item);
+          return this.store.put(itemId, item);
         });
       }
     } else if (event.updateTag) {
@@ -146,14 +146,15 @@ class ListingManager extends PublicObjectManager<Item> {
       if (ut.addItemId) {
         const itemId = bytesToHex(ut.addItemId);
         const item = await this.store.get(itemId);
-        item.tags = [...item.tags, tagId];
-        await this.store.put(itemId, item);
+        item.tags.push(tagId);
+        return this.store.put(itemId, item);
       }
       if (ut.removeItemId) {
         const itemId = bytesToHex(ut.removeItemId);
         const item = await this.store.get(itemId);
+        // remove `tagId` from item.tags array
         item.tags = [...item.tags.filter((id: `0x${string}`) => id !== tagId)];
-        await this.store.put(itemId, item);
+        return this.store.put(itemId, item);
       }
     }
   }
@@ -165,7 +166,7 @@ class ListingManager extends PublicObjectManager<Item> {
         metadata: new TextEncoder().encode(JSON.stringify(item.metadata)),
       }),
     );
-    await this.store.put(id, item);
+    return this.store.put(id, item);
   }
   //update argument passed here will only contain the fields to update.
   async update(update: Partial<Item>) {
@@ -197,10 +198,11 @@ class ListingManager extends PublicObjectManager<Item> {
         });
         item.quantity = update.quantity;
       }
-      await this.store.put(update.id!, item);
+      return item;
     };
 
-    Promise.all([p1, p2]);
+    const [_, item] = await Promise.all([p1, p2()]);
+    return this.store.put(update.id!, item);
   }
 
   getIterator() {
@@ -231,7 +233,7 @@ class ShopManifest extends PublicObjectManager<
         publishedTagId: bytesToHex(sm.publishedTagId),
         description: sm.description,
       };
-      await this.store.put("shopManifest", manifest);
+      return this.store.put("shopManifest", manifest);
     } else if (event.updateShopManifest) {
       const um = event.updateShopManifest;
       const manifest = await this.store.get("shopManifest");
@@ -256,7 +258,7 @@ class ShopManifest extends PublicObjectManager<
           chainId: um.chainId,
         };
       }
-      await this.store.put("shopManifest", manifest);
+      return this.store.put("shopManifest", manifest);
     }
   }
   async create(manifest: IShopManifest, shopId: `0x${string}`) {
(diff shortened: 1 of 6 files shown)


packages/stateManager/index.ts
```diff
@@ -1,104 +1,478 @@
 import { EventEmitter } from "events";
 import { RelayClient } from "@massmarket/client";
 import schema from "@massmarket/schema";
-
+import { bytesToHex, hexToBytes } from "viem";
+import { Uint8ArrayToParsedJSON } from "@massmarket/utils";
```
jchoi330, line 5: bufferToJSON

packages/stateManager/index.ts
```diff
@@ -1,104 +1,478 @@
...
+      const ci = event.createItem;
+      const id = bytesToHex(ci.eventId);
+      const item = {
+        id,
+        price: ci.price,
+        metadata: Uint8ArrayToParsedJSON(ci.metadata),
+        tags: [],
+        quantity: 0,
+      };
+      await this.store.put(id, item);
```
wanderer, line 118:
```suggestion
      return this.store.put(id, item);
```

packages/stateManager/index.ts
```diff
@@ -1,104 +1,478 @@
...
+      const ui = event.updateItem;
+      const id = bytesToHex(ui.itemId);
+      const item = await this.store.get(id);
+      if (ui.metadata) {
+        item.metadata = Uint8ArrayToParsedJSON(ui.metadata);
+      }
+      if (ui.price) {
+        item.price = ui.price;
+      }
+      await this.store.put(id, item);
```
wanderer, line 129:
```suggestion
      return this.store.put(id, item);
```

packages/stateManager/index.ts
```diff
@@ -1,104 +1,478 @@
...
+    } else if (event.changeStock) {
+      const cs = event.changeStock;
+      if (cs.itemIds) {
+        cs.itemIds.forEach(async (id: Uint8Array, i: number) => {
+          const itemId = bytesToHex(id);
+          const item = await this.store.get(itemId);
+          const diff = cs.diffs ? cs.diffs[i] : 0;
+          item.quantity = item.quantity + diff;
+          await this.store.put(itemId, item);
```
wanderer, line 138:
```suggestion
          return this.store.put(itemId, item);
```

packages/stateManager/index.ts
```diff
@@ -1,104 +1,478 @@
...
+        const itemId = bytesToHex(ut.addItemId);
+        const item = await this.store.get(itemId);
+        item.tags = [...item.tags, tagId];
+        await this.store.put(itemId, item);
+      }
+      if (ut.removeItemId) {
+        const itemId = bytesToHex(ut.removeItemId);
+        const item = await this.store.get(itemId);
+        item.tags = [...item.tags.filter((id: `0x${string}`) => id !== tagId)];
+        await this.store.put(itemId, item);
```
wanderer, line 156:
```suggestion
        return this.store.put(itemId, item);
```

packages/stateManager/index.ts
```diff
@@ -1,104 +1,478 @@
...
+      }
+    } else if (event.updateTag) {
+      // Add or remove tagId to item
+      const ut = event.updateTag;
+      const tagId = bytesToHex(ut.tagId);
+
+      if (ut.addItemId) {
+        const itemId = bytesToHex(ut.addItemId);
+        const item = await this.store.get(itemId);
+        item.tags = [...item.tags, tagId];
```
wanderer, line 149:
```suggestion
        item.push(tagId);
```

packages/stateManager/index.ts
```diff
@@ -1,104 +1,478 @@
...
+        await this.store.put(itemId, item);
+      }
+      if (ut.removeItemId) {
+        const itemId = bytesToHex(ut.removeItemId);
+        const item = await this.store.get(itemId);
+        item.tags = [...item.tags.filter((id: `0x${string}`) => id !== tagId)];
```
wanderer, line 155:
```suggestion
        // remove `tagId` from item.tags array
        item.tags = [...item.tags.filter((id: `0x${string}`) => id !== tagId)];
```

packages/stateManager/index.ts
```diff
@@ -1,104 +1,478 @@
...
+      }
+      if (update.quantity) {
+        const diff = update.quantity - item.quantity;
+        await this.client.changeStock({
+          itemIds: [hexToBytes(update.id!)],
+          diffs: [diff],
+        });
+        item.quantity = update.quantity;
+      }
+      await this.store.put(update.id!, item);
```
wanderer, line 200: We want to wait to store the update until P1 is done. That way if there is an error with the relay, the client will have not yet updated its state.

packages/stateManager/index.ts
```diff
@@ -1,104 +1,478 @@
...
+    };
+
+    Promise.all([p1, p2]);
```
wanderer, line 203:
```suggestion
    await Promise.all([p1, p2]);
    return this.store.put(update.id!, item);
```

packages/stateManager/index.ts
```diff
@@ -1,104 +1,478 @@
...
+  }
+
+  getIterator() {
```
wanderer, line 206:
```suggestion
  get iterator() {
```

packages/stateManager/index.ts
wanderer: Here we are always using `return` at the end of a control flow branch. This allows the reader of the code to know explicitly where the function stops executing.

packages/stateManager/index.ts
```diff
@@ -1,104 +1,478 @@
...
+  }
+
+  async create(item: Item) {
+    const id = bytesToHex(
+      await this.client.createItem({
+        price: item.price,
+        metadata: new TextEncoder().encode(JSON.stringify(item.metadata)),
+      }),
+    );
+    await this.store.put(id, item);
```
wanderer, line 168:
```suggestion
    return this.store.put(id, item);
```

packages/stateManager/index.ts
```diff
@@ -1,104 +1,478 @@
...
+    const ui: schema.IUpdateItem = {
+      itemId: hexToBytes(update.id!),
+    };
+    if (update.price) {
+      ui.price = update.price;
+    }
+    if (update.metadata) {
+      ui.metadata = new TextEncoder().encode(JSON.stringify(update.metadata));
+    }
+    const p1 = this.client.updateItem(ui);
```
wanderer, line 182:
```suggestion
    const opQueue = [this.client.updateItem(ui)];
```

packages/stateManager/index.ts
```diff
@@ -1,104 +1,478 @@
...
+      const item = await this.store.get(update.id!);
+      if (update.price) {
+        item.price = update.price;
+      }
+      if (update.metadata) {
+        item.metadata = update.metadata;
+      }
+      if (update.quantity) {
+        const diff = update.quantity - item.quantity;
+        await this.client.changeStock({
```
wanderer, line 194: This should be pulled out of the current closure and pushed to the opQeueu
```suggestion
       opQueue.push(this.client.changeStock({
```

packages/stateManager/index.ts
wanderer, line 203:
```suggestion
    Promise.all(opQueue);
```

packages/stateManager/index.ts
```diff
@@ -1,104 +1,478 @@
...
+      itemId: hexToBytes(update.id!),
+    };
+    if (update.price) {
+      ui.price = update.price;
+    }
+    if (update.metadata) {
+      ui.metadata = new TextEncoder().encode(JSON.stringify(update.metadata));
+    }
+    const p1 = this.client.updateItem(ui);
+    const p2 = async () => {
```
wanderer, line 183: change to
```suggestion
    opQueue.push(this.store.get(update.id!).then(item => {
```

wanderer (changes requested):

packages/stateManager/index.ts
jchoi330: i think the changestock promise has to be a part of whatever closure `const item = await this.store.get(update.id!);` is in because we have to calculate the diff

packages/stateManager/index.ts
```diff
@@ -1,104 +1,480 @@
...
+      if (cs.orderId.byteLength && cs.txHash) {
+        const orderId = bytesToHex(cs.orderId);
+        const order = await this.store.get(orderId);
+        order.status = Status.Complete;
+        order.txHash = bytesToHex(cs.txHash);
+        return this.store.put(orderId, order);
+      }
+    }
+  }
+  getIterator() {
```
wanderer, line 377:
```suggestion
  get iterator() {
```

packages/stateManager/index.ts
wanderer: actully you probably can move this to the abstract class

PR #83, opened 2024-07-24
Stream

jchoi330:


packages/client/test/stream.test.ts
```diff
@@ -66,6 +79,64 @@ describe("Stream", async () => {
     }
   });
 
+  test("Multiple Stream Creation", async () => {
```
wanderer, line 82: update the test description, we are not creating multiple streams.

packages/client/test/stream.test.ts
```diff
@@ -2,9 +2,22 @@ import { describe, assert, expect, test } from "vitest";
 import { ReadableEventStream } from "../src/stream";
 import { privateKeyToAccount } from "viem/accounts";
 import { hexToBytes } from "viem";
+import testVectorsData from "./testVectors.json" with { type: "json" };
 
 import schema, { PBObject, PBMessage, PBInstance } from "@massmarket/schema";
 
+let reqIdCounter: number = 1;
+function sequentialReqId() {
```
wanderer, line 10: why do we need this function? Why can we use a random id?

packages/client/test/stream.test.ts
```diff
@@ -66,6 +79,64 @@ describe("Stream", async () => {
...
+      events,
+    };
+    const client = new MockClient();
+    const stream = new ReadableEventStream(client);
+    stream.enqueue(pushEvent);
+    let count = 0;
+    for await (const evt of stream.stream) {
+      count++;
+      if (count === pushEvent.events.length) break;
+    }
```
wanderer, line 108: add expect(count).toEqual(...

wanderer (changes requested):

packages/client/test/stream.test.ts
cryptix: that doesn't break/cancel the draining of the stream though

packages/client/test/stream.test.ts
```diff
@@ -66,6 +76,70 @@ describe("Stream", async () => {
...
+    const stream = new ReadableEventStream(client);
+    stream.enqueue(pushReq);
+    // TODO: we need a way to close the stream once all requests have been pushed
+    let count = 0;
+    for await (const evt of stream.stream) {
+      count++;
+      // this might never happen if events are omitted by ReadableEventStream
+      // TODO: this might also exit to early if events would be duplicated...
+      if (count === pushReq.events.length) break;
+    }
```
cryptix, line 136: @wanderer do you have an idea how we might do this better?

packages/client/test/stream.test.ts
```diff
@@ -66,6 +76,70 @@ describe("Stream", async () => {
...
+    expect(count).toEqual(testVectors.events.length);
+  });
+
+  // TODO: should add version of the test above that creates random chunking of the test vectors
+  // and assert that they are all returned in order as individual events
```
cryptix, line 141: @jchoi330 you might want to try taking a stab at this but i think it can wait until a follow-up PR

cryptix (approved):
@wanderer unless you have a better idea on how to do the "is the stream drained completely" in packages/client/test/stream.test.ts i think we should merge this.

PR #84, opened 2024-07-26
Statemanager tests

jchoi330:


packages/client/src/index.ts
```diff
@@ -43,6 +43,9 @@ export class RelayClient extends EventEmitter {
   createEventStream() {
     return this.eventStream.stream;
   }
+  enqueue(pushReq: schema.EventPushRequest) {
```
wanderer, line 46: where is this being used?

.gitignore
```diff
@@ -36,3 +36,6 @@ dist-ssr
 # next
 **/.next
 packages/frontend/out
+
+# test Level DB
+packages/stateManager/tmpDb
```
wanderer, line 41: use [memdb](https://github.com/Level/memdown) so that it will not create this folder in the first place

packages/stateManager/index.ts
```diff
@@ -1,105 +1,537 @@
...
   }
-  abstract _processEvent(event: schema.ShopEvents): Promise<boolean>;
-  abstract create(obj: T): Promise<[any, any]>;
-  abstract update(obj: T): Promise<[any, any]>;
-  get(id: string): Promise<T> {
-    return this.db.get(id);
+  abstract _processEvent(event: schema.ShopEvents): Promise<void>;
+  abstract get(key?: string): Promise<T>;
+  get iterator() {
+    return () => {
```
wanderer, line 97: why return a function?

packages/stateManager/tests/stateManager.test.ts
```diff
@@ -0,0 +1,274 @@
...
+  );
+  let r = 0;
+  client.on("shopManifest", (e) => {
+    console.log("event:shopManifest", e);
+    r = 1;
+  });
+  client.on("updateShopManifest", (e) => {
+    console.log("event:updateShopManifest", e);
+  });
+  await vi.waitUntil(async () => {
```
jchoi330, line 44: trying to do something like this so we only move on to the tests once the event is received, but this waitUntil never resolves.

packages/stateManager/tests/stateManager.test.ts
```diff
@@ -0,0 +1,274 @@
...
+    return r === 1;
+  });
+  client.connect();
```
jchoi330, line 47: the event listeners need to be set up before we start send the vector events

packages/stateManager/index.ts
```diff
@@ -1,105 +1,475 @@
...
-    if (typeof event === schema.ShopEvents.UpdateItem) {
-      // load the item from the store
-      const item = await this.db.get(event.eventId);
-      Object.assign(item, event);
-      await this.db.put(event.eventId, item);
-      this.emit("update", item);
-      return true;
-    } else {
-      return false;
+class ShopManifest extends PublicObjectManager<
```
wanderer, line 199:
```suggestion
class ShopManifestManager extends PublicObjectManager<
```

packages/stateManager/index.ts
```diff
@@ -1,105 +1,475 @@
...
+      ui.price = update.price;
+    }
+    if (update.metadata) {
+      ui.metadata = convertToBuffer(update.metadata);
+    }
+    await this.client.updateItem(ui);
+  }
+
+  async changeStock(itemIds: `0x${string}`[], diffs: number[]) {
+    await this.client.changeStock({
```
wanderer, line 189:
```suggestion
    return this.client.changeStock({
```

packages/stateManager/index.ts
```diff
@@ -1,105 +1,475 @@
...
+  | (IShopManifest & UpdateShopManifest);
 
-// This is a an interface that is used to retrieve and store objects from a persistant layer
+// This is an interface that is used to retrieve and store objects from a persistant layer
 type Store<T extends ShopObjectTypes> = {
   put(key: string, value: T): Promise<void>;
   get(key: string): Promise<T>;
+  iterator: {
+    [Symbol.asyncIterator](): AsyncIterableIterator<T>;
+  };
```
wanderer, line 82:
```suggestion
   iterator (): Iterator
```

packages/stateManager/index.ts
```diff
@@ -1,105 +1,475 @@
...
-  get(id: string): Promise<T> {
-    return this.db.get(id);
+  abstract _processEvent(event: schema.ShopEvents): Promise<void>;
+  abstract get(key?: string): Promise<T>;
+  get iterator() {
+    return () => {
+      // @ts-expect-error FIXME
+      return this.store.iterator();
+    };
+  }
```
wanderer, line 101:
```suggestion
  get iterator() {
          return this.store.iterator();
  }
```

.gitignore
wanderer, line 41:
```suggestion

```

packages/client/src/index.ts
```diff
@@ -107,6 +107,7 @@ export class RelayClient extends EventEmitter {
     const eventWriteRequest = {
       event: signedEvent,
     };
+    this.eventStream.enqueue({ event: shopEvent });
```
wanderer, line 110: we should add a tests in the client to show that we are getting the event from the eventStream that we are writing.

packages/stateManager/package.json
```diff
@@ -1,19 +1,34 @@
...
-    "@massmarket/tsconfig": "workspace:*",
     "@massmarket/schema": "workspace:*",
-    "abstract-level": "^2.0.0",
-    "level": "^8.0.1"
+    "@massmarket/tsconfig": "workspace:*",
+    "@massmarket/utils": "workspace:*",
+    "events": "^3.3.0",
+    "level": "^8.0.1",
+    "viem": "^2.16.3",
+    "vitest": "^2.0.4"
```
wanderer, line 23: `vitest` should be in devDep. Please check the rest to make sure that they are in the correct place

packages/stateManager/tests/mockClient.ts
```diff
@@ -0,0 +1,193 @@
...
+    };
+  };
+};
+
+export type IncomingEvent = {
+  request: schema.EventPushRequest;
+  done: () => void;
+};
+
+export class MockClient extends EventEmitter {
```
wanderer, line 90: Ideally MockClient would implement the interface provided by RelayClient. This way we have some garrentees that the methods are in sync. 
```suggestion
interface IRelayClient extends RelayClient {
}

export class MockClient extends EventEmitter implements IRelayClient {
```

packages/stateManager/tests/mockClient.ts
```diff
@@ -0,0 +1,193 @@
+// SPDX-FileCopyrightText: 2024 Mass Labs
+//
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+import { EventEmitter } from "events";
```
wanderer, line 5:
```suggestion
```

packages/stateManager/tests/stateManager.test.ts
```diff
@@ -0,0 +1,270 @@
+import { describe, expect, test, vi } from "vitest";
+import { Level } from "level";
+import { StateManager } from "@massmarket/stateManager";
+import { testVectors } from "@massmarket/schema";
+import { MockClient } from "./mockClient";
+import { randomAddress, random32BytesHex } from "@massmarket/utils";
+
+export const zeroAddress = "0x0000000000000000000000000000000000000000";
```
wanderer, line 8: move to utils

wanderer, line 8: move to utils
```suggestion

```

packages/stateManager/tests/stateManager.test.ts
```diff
@@ -0,0 +1,341 @@
...
+    const shop = await stateManager.shopDetails.get();
+    expect(shop.name).toEqual("Test Shop");
+    expect(shop.description).toEqual("Testing shopManifest");
+  });
+
+  test("ShopManifest - update", async () => {
+    let r = 0;
+    client.on("updateShopManifest", async () => {
+      r = 1;
+    });
```
wanderer, line 183:
```suggestion
const manifestUpdateComplete = new Promise((resolve, reject) => {
    stateManager.manifest.on("update", resolve);
})

```

packages/stateManager/tests/stateManager.test.ts
```diff
@@ -0,0 +1,341 @@
...
+        },
+      ],
+      setBaseCurrency: {
+        chainId: 1,
+        tokenAddr: zeroAddress,
+      },
+    });
+    await vi.waitUntil(async () => {
+      return r === 1;
+    });
```
wanderer, line 202: await manifestUpdateComplete;

packages/stateManager/tests/stateManager.test.ts
```diff
@@ -0,0 +1,341 @@
...
+    const shop = await stateManager.shopDetails.get();
+    expect(shop.name).toEqual("Test Shop");
+    expect(shop.description).toEqual("Testing shopManifest");
+  });
+
+  test("ShopManifest - update", async () => {
+    let r = 0;
```
wanderer, line 180:
```suggestion
```

packages/stateManager/tests/stateManager.test.ts
```diff
@@ -0,0 +1,341 @@
...
+    client.on("updateShopManifest", async () => {
+      r = 1;
+    });
+    await stateManager.shopDetails.update({
```
wanderer, line 184:
```suggestion
    await stateManager.manifest.update({
```

packages/stateManager/index.ts
```diff
@@ -1,104 +1,548 @@
...
+      } else if (key === "publishedTagId") {
+        updateShopManifest[key] = hexToBytes(updateShopManifest.publishedTagId);
+      } else if (key === "setBaseCurrency") {
+        updateShopManifest[key] = {
+          tokenAddr: hexToBytes(updateShopManifest[key].tokenAddr),
+          chainId: updateShopManifest[key].chainId,
+        };
+      }
+    }
+    return this.client.updateShopManifest(updateShopManifest);
```
wanderer, line 306:
```suggestion
    const id = await this.client.updateShopManifest(updateShopManifest);
    return new Promise((resolve, reject) => {
       function onUpdate(update){
           if(update.id ===  id) {
             resolve();
             this.removeListener('update', onUpdate)
            } 
       }
       this.on('update', onUpdate)
    })
```

packages/stateManager/index.ts
```diff
@@ -1,104 +1,548 @@
...
+        ...manifest,
+        publishedTagId: new Uint8Array(32),
+        profilePictureUrl: "https://http.cat/images/200.jpg",
+      },
+      shopId,
+    );
+    return bytesToHex(id);
+  }
+
+  async update(um: Partial<ShopManifest>) {
```
wanderer, line 284:
```suggestion
// this function resolve once the update has been accepted by the relay and written to the database. This will happen after the `update` event has been fired. 
  async update(um: Partial<ShopManifest>) {
```

wanderer (changes requested):

packages/client/src/index.ts
```diff
@@ -107,6 +107,11 @@ export class RelayClient extends EventEmitter {
     const eventWriteRequest = {
       event: signedEvent,
     };
+    const pushReq = new schema.EventPushRequest({
+      requestId: requestId(),
+      events: [signedEvent],
+    });
+    this.eventStream.enqueue(pushReq);
     return this.encodeAndSend(schema.EventWriteRequest, eventWriteRequest);
```
wanderer, line 115:
```suggestion
    // Wait for the relay to response to our write before enqueing the event locally 
    await this.encodeAndSend(schema.EventWriteRequest, eventWriteRequest);
    this.eventStream.enqueue(pushReq);
```

packages/stateManager/index.ts
```diff
@@ -1,104 +1,548 @@
...
     }
   }
-  create(item: Item) {
-    return Promise.all([
-      this.db.put(item.id, item),
-      this.client.createItem(item),
-    ]);
+  async create(name: string) {
+    const eventId = await this.client.createTag({ name });
+    return bytesToHex(eventId);
```
wanderer, line 477: Why are we return the eventId here?

packages/stateManager/index.ts
```diff
@@ -1,104 +1,548 @@
...
+        profilePictureUrl: "https://http.cat/images/200.jpg",
+      },
+      shopId,
+    );
+    return bytesToHex(id);
+  }
+
+  async update(um: Partial<ShopManifest>) {
+    //Convert tokenAddr and publishedTagId to bytes before sending to client.
+    const updateShopManifest: schema.IUpdateShopManifest = { ...um };
```
wanderer, line 286: Having some comment explaining the rations this line would be helpful

packages/stateManager/index.ts
```diff
@@ -1,104 +1,636 @@
...
+    } else if (event.changeStock) {
+      const cs = event.changeStock;
+      if (cs.itemIds) {
+        await Promise.all(
+          cs.itemIds.map(async (id: Uint8Array, i: number) => {
+            const itemId = bytesToHex(id);
+            const item = await this.store.get(itemId);
+            const diff = cs.diffs ? cs.diffs[i] : 0;
+            item.quantity = item.quantity + diff;
+            await this.store.put(itemId, item);
```
cryptix, line 143:
```suggestion
            return this.store.put(itemId, item);
```

we need to return the promise here so that it ends up in the list for `Promise.all`.

packages/stateManager/index.ts
```diff
@@ -1,104 +1,636 @@
...
+}
+interface ShopCurrencies {
+  tokenAddr: `0x${string}`;
+  chainId: number;
+}
+interface CreateShopManifest {
+  name: string;
+  description: string;
+}
+type ShopManifest = CreateShopManifest & {
```
cryptix, line 63: why do we have both here? 🤔

packages/stateManager/index.ts
```diff
@@ -1,104 +1,636 @@
...
+  async create(item: Partial<Item>): Promise<Item> {
+    const eventId = await this.client.createItem({
+      price: item.price,
+      metadata: convertToBuffer(item.metadata),
+    });
+
+    return new Promise((resolve, reject) => {
+      const remove = () => this.removeListener("createItem", onCreate);
+      function onCreate(create: Item) {
+        if (create.id === bytesToHex(eventId)) {
```
cryptix, line 183: it isn't immediately obvious why we compare against `create.id` here, since that field doesn't exist on the network/protobuf level.

in general i'd advice to keep the name for the ID-field every wehre the same, as much as possible.

Then we should also be possible to turn this promise-waits-for-event construction into a  function that is re-usable across stores, by only passing in the event type and id in.

packages/stateManager/index.ts
```diff
@@ -1,104 +1,636 @@
...
-  update(item: Item) {
-    return Promise.all([
-      // need to get the item and merge it before storing
-      this.db.put(item.id, item),
-      this.client.createItem(item),
-    ]);
+    return new Promise((resolve, reject) => {
+      const remove = () => this.removeListener("updateItem", onUpdate);
+      function onUpdate(item: Item) {
+        if (update.id === item.id) {
```
cryptix, line 208: this is deceptively working but incorrect.

you need to be careful here to not unblock i.e. the UI calling `.update()` via a network request also updating the same object. In short, this needs to use the `eventId` not the `update.itemId`.

Just to elaborate the bug a bit:

- the UI could call `stateManager.update(itemId=23, price=$somePrice)`, the network client will create an eventId=123 for the update that goes out to the network momentarily.
- another client/clerk might have updated that same item just before
- the stateManager receives `updateItem(eventId=666, itemId=23, title=$newTitle)`

now, if you listen for the itemId instead of the eventId, you unblock for the wrong update and have a gnarly concurrency bug on your hand.

packages/stateManager/index.ts
```diff
@@ -1,104 +1,636 @@
...
+        ...manifest,
+        publishedTagId: new Uint8Array(32),
+        profilePictureUrl: "https://http.cat/images/200.jpg",
+      },
+      shopId,
+    );
+    return new Promise((resolve, reject) => {
+      const remove = () => this.removeListener("createShopManifest", onUpdate);
+      function onUpdate(create: ShopManifest) {
+        if (create.tokenId === shopId) {
```
cryptix, line 323: same as above.

you need to use the eventId that you get from the network client and also propagate that back through the emitter.

cryptix (changes requested):
really close but there is a bug on the promise<>eventemitter thing.

packages/stateManager/tests/stateManager.test.ts
```diff
@@ -0,0 +1,343 @@
...
+    const listingIterator = stateManager.items.iterator;
+    for await (const [id, item] of listingIterator) {
+      const vectorItem = vectorState.items[id];
+      expect(item.price).toEqual(vectorItem.price);
+      const parsed = JSON.parse(vectorItem.metadata);
+      expect(item.metadata.title).toEqual(parsed.title);
+      expect(item.metadata.image).toEqual(parsed.image);
+      expect(item.quantity).toEqual(vectorItem.stock_qty);
+    }
+  });
```
cryptix, line 120: also check we iterated over all items after the for await {} block.

cryptix: same for tags

packages/utils/index.ts
```diff
@@ -31,3 +31,13 @@ export function hexToBase64(hex: string) {
   const u8 = new Uint8Array(toBytes(hex));
   return Buffer.from(u8).toString("base64");
 }
+
+export const bufferToJSON = (metadata: Uint8Array) => {
+  return JSON.parse(new TextDecoder().decode(metadata));
+};
+
+export const convertToBuffer = (data: any) => {
```
cryptix, line 39:
```suggestion
export const stringifyToBuffer = (data: any) => {
```

cryptix (commented):
A few small nitpicks but i think we are really close now!

packages/stateManager/tests/stateManager.test.ts
```diff
@@ -0,0 +1,343 @@
...
+      expect(o.status).toEqual("PENDING");
+      expect(o.items).toEqual(order.items);
+    }
+  });
+});
+
+describe("CRUD functions update stores", async () => {
+  const client = new MockClient();
+  const stateManager = new StateManager(
+    // @ts-expect-error FIXME looking for client to be RelayClient
```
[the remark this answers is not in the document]
jchoi330: we can redefine what the stateManager takes as client type to accommodate both MockClient and RelayClient. Especially since the state manager also does not utilise all the relay client functions

PR #85, opened 2024-08-02
sign-in with ethereum

cryptix:


cryptix:
hmm.. one thing we need to fix:

Screenshot 2024-08-02 at 14 03 51

the URI / domain is supposed to be the _shop url_ not the _relay_ we are signing in to. Otherwise it's flagged as a phishing attempt. Need to mull this one over.

cryptix:
@jchoi330 just flagged you FYI. i will take care of the phishing thing.

packages/frontend/src/app/components/login/MerchantConnectWallet.tsx
```diff
@@ -113,6 +113,7 @@ const MerchantConnectWallet = ({ close }: { close: () => void }) => {
             clientWallet,
             false,
             shopId,
+            new URL(window.location.href),
```
cryptix, line 116: @wanderer ended up doing this to fix the phishing warning. Not sure if this is the best way of going about it but it works.

PR #87, opened 2024-08-07
[WIP PR] frontend x stateManager

jchoi330:
1. add test for addPayee in updateShpManifest
2. cleanup interface names 
3. bring statemanager into frontend workspace


packages/stateManager/index.ts
```diff
@@ -330,6 +329,29 @@ class ShopManifestManager extends PublicObjectManager<ShopManifest> {
         }
         manifest.addAcceptedCurrencies = filtered;
       }
+      if (um.addPayee) {
```
cryptix, line 332: * **note to self** do we have removePayee in the vectors?

packages/stateManager/tests/stateManager.test.ts
```diff
@@ -231,6 +231,12 @@ describe("CRUD functions update stores", async () => {
         chainId: 1,
         tokenAddr: zeroAddress,
       },
```
cryptix, line 233: these updates should be separate sub-tests so we can see which/where it fails but as discussed, separate PR is 👌🏻

packages/frontend/src/types/index.ts
```diff
@@ -6,22 +6,25 @@ export type ItemId = `0x${string}`;
 export type TagId = `0x${string}`;
 export type OrderId = `0x${string}`;
 
-export type IProduct = {
-  id: ItemId;
+export interface Item {
```
cryptix, line 9: These should be removed in favour of the definitions in the state manager

https://github.com/masslbs/Tennessine/blob/a57bdd9f91daa81fd3ded1ceb60dc3e1e8c119fd/packages/stateManager/index.ts#L29-L86

cryptix (changes requested):
let's:

* split up / substitute usage of types in frontend with the ones in state manager
* if you remove buildState also remove the reducers
* smol nitpick about missing tests for remove currency and payee but can be in separate PR

packages/stateManager/index.ts
cryptix: we don't. we only have `remove_accepted_currencies` right now.

cryptix: but let's add crud tests for

* removing a payee that doesn't exist (should throw an _not found_ error)
* add a payee with the same name (should throw an _already exists_ error)

the relay would _also_ tell you this but i think it would be nicer to catch this before sending a request over the network.

cryptix (commented):
so much deleted code..! 🤩 

the useeffect-if-statemanger-thing is not a blocker from my end but might be cool to find a short-hand and-or different way to make sure stateManager is always initialized

packages/frontend/src/app/components/checkout/CartButton.tsx
```diff
@@ -2,16 +2,27 @@
...
 const CartButton = () => {
-  const { orderItems, orderId } = useStoreContext();
+  const { stateManager, getOrderId } = useStoreContext();
+  const [cartItemIds, setItemIds] = useState<ItemState | null>(null);
 
-  const activeCartItems = orderId && orderItems.get(orderId)?.items;
-  const arr = activeCartItems ? Object.values(activeCartItems) : [];
+  useEffect(() => {
+    (async () => {
+      if (stateManager) {
```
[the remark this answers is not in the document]
jchoi330: we could instantiate the stateManager with an initial loading value like LoadingStateManager instead of with null. This way we wont have to check whether stateManager is truthy everytime. We would just have to return a new promise with initial values for each state, manifest, products, orders, tags, etc.


```
class LoadingManifestManager<T extends ShopObjectTypes> extends EventEmitter {
  constructor() {
    super();
  }
  async get(key?: string): Promise<T> {
    return new Promise(() => {
      return { name: "", publishedTagId: null };
    });
  }
  async iterator() {
    return;
  }
}
export class LoadingStateManager {
  readonly items;
  readonly tags;
  readonly manifest;
  readonly orders;
  readonly keycards;
  constructor() {
    this.items = new LoadingManager();
    this.tags = new LoadingManager();
    this.manifest = new LoadingManifestManager();
    this.orders = new LoadingManager();
    this.keycards = new LoadingManager();
  }
}
```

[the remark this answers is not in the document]
jchoi330: @cryptix okay I implemented the initial loading state for statemanager. take a look at commit [0e5b4a0]

packages/frontend/src/context/StoreContext.tsx
```diff
@@ -2,678 +2,156 @@
...
+    if (openOrders && openOrders.length) {
+      order_id = openOrders[0] as OrderId;
+      setOrderId(order_id);
+    } else {
+      // If open order id does not exist, create new id
+      const { id } = await stateManager!.orders.create();
+      setOrderId(id);
+      order_id = id;
+    }
+    return order_id;
```
jchoi330, line 142: I think orderId should live globally (in context) since we are going to have to map the open orders, get the signed address for the order, then key in that address in the KeyCard store.

packages/frontend/src/app/cart/NewCart.tsx
```diff
@@ -7,40 +7,52 @@ import React, { useState, useEffect, useRef } from "react";
...
-    if (orderId) {
-      const items = orderItems.get(orderId)?.items || null;
-      setActiveCartItems(items);
-    }
-  }, [orderId, orderItems]);
 
-  const noItems =
-    !orderId || !activeCartItems || !Object.keys(activeCartItems).length;
+  useEffect(() => {
+    (async () => {
```
wanderer, line 31: Does the following work?
```suggestion
  useEffect(async () => {
```
its looks like you have a unnecessarily nested function here

packages/frontend/src/app/components/checkout/CartButton.tsx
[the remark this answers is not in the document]
wanderer: @cryptix @jchoi330 I'm not sure what the best way to do this is either

packages/frontend/src/context/StoreContext.tsx
```diff
@@ -2,678 +2,156 @@
...
+        const db = new Level(`./${dbName}`, {
+          valueEncoding: "json",
         });
-      });
-    })();
-  }, [acceptedCurrencies]);
-
-  useEffect(() => {
-    (async () => {
-      const { Level } = await import("level");
```
jchoi330, line 127: document why are importing level like this

PR #89, opened 2024-08-12
adding order Statuses as key to orderStore

jchoi330:
Creates keys in OrderManager for pending/canceled/complete with values orderIds[]

### DIFF
--- a/packages/stateManager/index.ts
+++ b/packages/stateManager/index.ts
@@ -44,7 +44,7 @@ export interface Tag {
 }
 
 export type KeyCard = `0x${string}`;
-enum Status {
+export enum Status {
   Failed = "FAILED",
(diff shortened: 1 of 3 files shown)


packages/stateManager/index.ts
```diff
@@ -424,12 +436,27 @@ class OrderManager extends PublicObjectManager<Order> {
...
+      } catch (error) {
+        const e = error as Error;
+        if (e.notFound) {
+          openOrders.push(id);
+        } else {
+          throw new Error(e.code);
+        }
+      }
+      await this.store.put(Status.Pending, openOrders);
+
```
wanderer, line 453: Since you use this pattern several times see if you can generalize it by making it a private function in this class.

PR #94, opened 2024-08-19
[WIP] UI Vitest

jchoi330:


packages/frontend/src/app/create-store/page.tsx
```diff
@@ -171,8 +163,16 @@ const StoreCreation = () => {
...
         const newPubId = await stateManager.tags.create("visible");
-        const path = await relayClient!.uploadBlob(avatar as FormData);
+        await stateManager.tags.create("remove");
+
+        //Testing dom does not support FormData and test client will fail with:
+        //Content-Type isn't multipart/form-data
+        //so if it is a test env, we are skipping uploadBlob
+        const path = process.env.TEST
+          ? { url: "/" }
+          : await relayClient!.uploadBlob(avatar as FormData);
```
cryptix, line 175: seen these in a few places. I think we should be able to do something about that. Not sure what yet though and think it's fine to clean up later.

packages/frontend/tests/CreateStore.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+    },
+  }));
+
+  vi.mock(import("wagmi"), async (importOriginal) => {
+    const mod = await importOriginal(); // type is inferred
+    return {
+      ...mod,
+      useWalletClient() {
+        const account = privateKeyToAccount(
+          "0x2a871d0798f97d79848a013d4936a73bf4cc922c825d33c1cf7073dff6d409c6",
```
cryptix, line 42: this should go into the `.env.development` file

packages/frontend/tests/CreateStore.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+      await user.click(saveBtn);
+    });
+    await waitFor(async () => {
+      expect(spy).toHaveBeenCalled();
+      const manifest = await sm.manifest.get();
+      expect(manifest.name).toEqual("New Store Name II");
+      expect(manifest.description).toEqual("New Store Description II");
+      expect(manifest.payee[0].addr).toEqual(payee);
+      expect(manifest.setBaseCurrency?.tokenAddr).toEqual(zeroAddress);
+    });
```
cryptix, line 144: i'm wondering if this is enough of check to see if we created a store but it might be.

just asking to understand whether the other steps to the process (mint store nft, upload metadata) are tested elsewhere.

cryptix (commented):
just a few minor nitpicks. 

looking good overall!

packages/frontend/tests/CreateStore.test.tsx
jchoi330: @cryptix this actually tests the minting! Because it has to mint the store and pass permission check in order to enroll the keycard, and then once kc is enrolled successfully it will create/update the shop manifest because of the conditionals in the create-store file

PR #98, opened 2024-08-26
convert to bigNum

jchoi330:


wanderer:
we need to use `[BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt)` not bigNumber. I'm pretty sure I said bignumber too unfortunately. Bignum is what we used to use before BigInt became standardized. But since BigInt is built in, it will be more performant and will not introduce another dependency.

PR #99, opened 2024-08-27
convert to bigint

jchoi330:


packages/frontend/src/app/components/transactions/QRScan.tsx
```diff
@@ -47,7 +47,7 @@ function QRScan({
           </div>
           <p className="text-primary-gray">scan to pay</p>
           <h2>
-            {erc20Amount} {symbol}
+            {Number(erc20Amount)} {symbol}
```
wanderer, line 50: Number(erc20Amount) will display the incorrect amount since it will over overflow. We shouldn't convert back into number and we should only truncate the number in on critical places. Here a user might try to copy the amount to pay manually.

wanderer (changes requested):

wanderer (approved):

PR #109, opened 2024-09-16
saving walletAddress as level key

jchoi330:


packages/frontend/src/context/StoreContext.tsx
```diff
@@ -84,55 +85,33 @@ export const StoreContextProvider = (
...
+      //Getting all public keys by current wallet address.
+      const publicKeys = await stateManager?.keycards.get(
+        walletAddress.toLowerCase() as `0x${string}`,
+      );
+      //Getting address from public key.
+      const addresses = publicKeys.map((a: OrderId) => {
+        return Address.fromPublicKey(hexToBytes(a)).toString();
+      });
+      for (const o of openOrders) {
+        //Go through the open orders and see if the signer address is one of the public keys created with the wallet address.
```
cryptix, line 102: we shouldn't do this again and again when loading an order.

ideally we just check the signature once when we receive an event from the relay / before persisting it in the state manager. Then we would know that all events in the state manager where valid at the time of storing them. 

(clerks that were removed from a store are a separate problem we are tackling later.)

cryptix (approved):
Goes in the right direction but we shouldn't special-case the verification just to orders.

I'd move the check before the persistence update in the state manager. Then we can check all events in one place. 

I'd leave it up to you if you want to do this in this PR or a follow-up. The keyCards store change look good to me.

jchoi330:
@cryptix For verifying all events, are we checking the event signature address against all public keys created regardless of wallet address?

packages/frontend/src/context/StoreContext.tsx
jchoi330: I added line 90  `if (orderId) return orderId;` so this fixes the issue of having to calculate the orderId associated with the wallet address every time. 

In regards to this suggestion: _ideally we just check the signature once when we receive an event from the relay / before persisting it in the state manager._
I'm not sure where we would check the signature before updating the db. I think this would require us passing the wallet address to the stateManager, and then we still have the issue of somehow being able to access the keycard Store inside orderManager in order to get all the public keys associated with the WA in order to filter the correct orderId.

wanderer: > I'm not sure where we would check the signature before updating the db.

we should do it here
https://github.com/masslbs/Tennessine/blob/6c680f8e0b6c30c0f182cec29136491d284a12e8/packages/stateManager/index.ts#L641
before we call _processEvent on the objectStores

PR #111, opened 2024-09-18
verifying all events

jchoi330:


packages/stateManager/index.ts
```diff
@@ -615,8 +627,45 @@ export class StateManager {
       this.keycards,
     ];
     const stream = this.client.createEventStream();
+    const publicClient = createPublicClient({
+      chain: hardhat,
+      transport: http(),
+    });
+
+    const count = (await publicClient.readContract({
```
jchoi330, line 635: _getRelayCount_ and _getRelayCount_ contract calls are not erroring out but it returns empty. Shouldn't it return my locally running address?

jchoi330: my deployment addresses seem to all match too

cryptix: By default the shop has zero relays. The function to to do this is called `addRelay`, and it takes the shop and the relay NFT id.

https://github.com/masslbs/pystoretest/blob/a9307bb29057fad4f5053ab2202f1cef8f3361b5/client.py#L296

for MVP i'd do this together with the shop mint

i introduced the relay nft id a while ago but we were only using it for the new enroll. It's bundled up in the RelayEndpoint type:
https://github.com/masslbs/Tennessine/blob/90c10f0f0baa6c0ae5a01283ea4a26741ce3f437/packages/client/src/index.ts#L28

packages/stateManager/index.ts
```diff
@@ -615,8 +627,45 @@ export class StateManager {
       this.keycards,
     ];
     const stream = this.client.createEventStream();
+    const publicClient = createPublicClient({
```
jchoi330, line 630: for dev use hardhat. on prod it will always be mainnet

packages/frontend/src/app/create-store/page.tsx
```diff
@@ -117,6 +118,12 @@ const StoreCreation = () => {
         });
 
         if (transaction!.status == "success") {
+          //Add relay tokenId
+          const relayURL =
+            (process && process.env["RELAY_ENDPOINT"]) ||
+            "ws://localhost:4444/v2";
+          const relayEndpoint = await discoverRelay(relayURL);
```
cryptix, line 125:
```suggestion
          // TODO: use process.env.NEXT_PUBLIC_RELAY_TOKEN_ID in prod
          const relayEndpoint = await discoverRelay(relayURL);
```


We should probably condense what i wrote here a bit more, since `discoverRelay` only works in testing.

https://github.com/masslbs/Tennessine/blob/90c10f0f0baa6c0ae5a01283ea4a26741ce3f437/packages/frontend/src/context/MyContext.tsx#L286-L292

packages/frontend/tests/ProductDetail.test.tsx
```diff
@@ -116,7 +119,11 @@ describe("Product Detail Component", async () => {
         name: /Im sure/i,
       });
       await user.click(confirmButton);
+    });
+
+    await waitFor(async () => {
       const res = await sm.items.get(itemId);
+      console.log({ res });
```
cryptix, line 126:
```suggestion
```

packages/frontend/tests/ProductDetail.test.tsx
```diff
@@ -78,10 +78,13 @@ describe("Product Detail Component", async () => {
       });
       await user.click(screen.getByTestId("addToCart"));
     });
-    // Check that the item (2qty) we added to cart above is saved in stateManager
-    const openOrder = await sm.orders.getStatus(Status.Pending);
-    const orderDetails = await sm.orders.get(openOrder[0]);
-    expect(orderDetails.items[itemId]).toEqual(2);
+
+    await waitFor(async () => {
```
cryptix, line 82: can you tell me why we need these `waitFor`s now?

packages/stateManager/index.ts
```diff
@@ -567,16 +570,23 @@ class KeyCardManager extends PublicObjectManager<KeyCard> {
...
+      await this.store.put(addressFromPubKey, userWalletAddr);
+      this.emit("newKeyCard", addressFromPubKey);
       return;
     }
   }
 
   get(key: `0x${string}`) {
     return this.store.get(key);
   }
+  addRelayAddress(key: `0x${string}`) {
```
cryptix, line 586: ah.. just to make the naming a bit clearer. I thought we might want to different functions (e.g. `addRelay` and `addKeyCard`) since we have two different paths (from the contract registries and from the event stream/account messages). But if the stateManager only persists the addresses, let's just call it `addAddress`.

put differently, keyCards are not relays 🙃

packages/stateManager/index.ts
```diff
@@ -615,10 +628,47 @@ export class StateManager {
       this.keycards,
     ];
     const stream = this.client.createEventStream();
+    const shopClient = createPublicClient({
+      chain: process.env.DEV ? hardhat : mainnet,
+      transport: http(),
+    });
+    //When we inititally create a shop, we are saving the relay tokenId => shopId.
+    //Here, we are retrieving all the relay addresses associated with the shopId and saving them to keycards store.
+    //Since some shopEvents are signed by a relay, we need to include these addresses when verifying the event signer.
```
cryptix, line 637: 👌🏻

packages/stateManager/index.ts
```diff
@@ -615,10 +628,47 @@ export class StateManager {
...
     for await (const event of stream) {
-      for (const storeObject of storeObjects) {
-        await storeObject._processEvent(event.event);
+      try {
+        const wa = await this.keycards.get(event.signer.toLowerCase());
+        for (const storeObject of storeObjects) {
+          await storeObject._processEvent(event.event);
+        }
+      } catch (error) {
+        console.log(`Event not verified: ${(event.signer, event.event)}`);
```
cryptix, line 671: hmmm.. these would vanish silently for most users but I'm not sure what exactly we _do_ want to happen here. cc @wanderer

cryptix (commented):
Awesome! A few nitpicks but nothing major

packages/blockchain/index.ts
```diff
@@ -26,7 +26,14 @@ export type WalletClientWithAccount = WalletClient<
 
 export class BlockchainClient {
   constructor(public shopId = bytesToHex(eventId())) {}
-
+  addRelay(wallet: WalletClientWithAccount, tokenId: Address) {
```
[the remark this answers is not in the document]
wanderer: this is confusing; Address is typed a `0x{string}` which will work for tokenID. And I don't think there is a way of specifing the string length. So we cannot differentiate between `Addresses` which are 20bytes and uint256 which are 32 bytes, but both can be represent as a hex string.  If it is more convient to use a hex string then we should just create a new type. Else go with bigint.

packages/stateManager/index.ts
```diff
@@ -615,10 +628,47 @@ export class StateManager {
...
+        await this.keycards.addRelayAddress(ownerAdd);
+      }
+    }
+
     //Each event will go through all the storeObjects and update the relevant stores.
     for await (const event of stream) {
-      for (const storeObject of storeObjects) {
-        await storeObject._processEvent(event.event);
+      try {
+        const wa = await this.keycards.get(event.signer.toLowerCase());
```
wanderer, line 666: unused var
```suggestion
        await this.keycards.get(event.signer.toLowerCase());
```

packages/stateManager/index.ts
wanderer: Maybe this should boil up to the UI? Which mean we would not catch it here.

wanderer (changes requested):
Looking good. We defaintly want some tests with regards to bad signature on events though. I think we probably should remove the catch in the stateManger, and add a catch in the frontend.

packages/stateManager/index.ts
jchoi330: remove the try catch and write test that halts the operation

packages/stateManager/index.ts
```diff
@@ -615,8 +628,41 @@ export class StateManager {
       this.keycards,
     ];
     const stream = this.client.createEventStream();
+    const shopClient = createPublicClient({
+      chain: process.env.DEV ? hardhat : mainnet,
+      transport: http(),
+    });
```
wanderer, line 634:
```suggestion
```
this should just be an argument in constructor. The reason is the most likely the frontend need to create a publicClient anyways (https://github.com/masslbs/Tennessine/blob/90c10f0f0baa6c0ae5a01283ea4a26741ce3f437/packages/frontend/src/context/MyContext.tsx#L127) and we will want to use that one instead of creating two instances

wanderer (changes requested):

PR #112, opened 2024-09-24
Rename listing

jchoi330:


packages/client/test/guestCheckout.test.ts
```diff
@@ -183,106 +231,89 @@ test("create and enroll guest", { timeout: 10000 }, async () => {
...
-
-test("single item checkout with a guest", { timeout: 10000 }, async () => {
-  // give the guest account some money to spend
-  const txHash = await wallet.writeContract({
-    address: abi.addresses.Eddies as Address,
-    abi: abi.Eddies,
-    functionName: "mint",
-    args: [guestAccount.address, 999999999999],
-  });
+// test("single item checkout with a guest", { timeout: 10000 }, async () => {
```
wanderer, line 249: instead of commenting this out, use `test.skip`

wanderer (changes requested):
Looks good, just need to use test.skip instead of commenting out the tests

packages/client/src/index.ts
```diff
@@ -5,19 +5,13 @@
 import { EventEmitter } from "events";
 
 import { WebSocket } from "isows";
-import { bytesToHex, hexToBytes, hexToBigInt, toBytes } from "viem";
+import { bytesToHex, hexToBytes, hexToBigInt, toBytes, toHex } from "viem";
```
wanderer, line 8: are we using `toHex` somewhere?

PR #113, opened 2024-09-26
frontend changes for v3

jchoi330:


packages/stateManager/tests/mockClient.ts
```diff
@@ -144,10 +143,21 @@ export class MockClient implements IRelayClient {
...
+  }
+  //Mimics client-fired event paymentDetails after commit event - for testing paymentDetails gets stored correctly in stateManager.
+  async sendPaymentDetails(orderId: `0x${string}`) {
+    return this.sendShopEvent({
       updateOrder: {
-        id: orderId,
+        id: hexToBytes(orderId),
         eventId: hexToBytes(
           "0x32b36377007de4ab0fcc3eabb1ef3a7096c42004c14babc3638f81b9d0982625",
         ),
```
cryptix, line 163:
```suggestion
```

eventId's are not thing anymore. Is probably ignored and thus not a real issue but might be confusing to stumble over in the future.

jchoi330: I know the relay ignores it but I am including eventId for the eventListenAndResolve function in stateManager. We could use id instead of eventId for the statemanager to listen and resolve, but in the edge case that two people update the same tag at the same time,  we don't want the update function to resolve from somebody else's update

packages/frontend/tests/Checkout.test.tsx
```diff
@@ -27,57 +27,57 @@ describe("Checkout", async () => {
   });
   let orderId: `0x${string}`;
 
-  test("Update shipping details", async () => {
+  test.only("Update shipping details", async () => {
```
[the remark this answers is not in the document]
jchoi330: we will have to ignore this whole file till orderHash stuff is resolved

PR #116, opened 2024-10-11
separate authenticate from connect in client

jchoi330:


packages/client/test/guestCheckout.test.ts
```diff
@@ -160,62 +162,6 @@ test("should create a item", { timeout: 10000 }, async () => {
   });
 });
 
-test("should update stock", { timeout: 10000 }, async () => {
```
jchoi330, line 163: this is moved to relayClient test

packages/client/test/relayClient.test.ts
```diff
@@ -402,30 +443,13 @@ describe("user behaviour", () => {
...
-      await relayClient2.updateShopManifest({
-        addPayee: {
-          address: { raw: randomBytes(20) },
-          callAsContract: false,
-          chainId: 1,
-          name: "new payee",
-        },
-      });
-    });
-    test("client 1 receives events from createEventStream", async () => {
```
jchoi330, line 418: we're testing this in the checkout tests above

wanderer (changes requested):

PR #121, opened 2024-10-24
Deno 4

wanderer:


wanderer:
` deno run --unstable-unsafe-proto build ` is need to build the frontend because of [this](https://github.com/denoland/deno/issues/25359#issuecomment-2437841392)

PR #125, opened 2024-11-06
Fix ci

cryptix:


cryptix:
https://github.com/masslbs/Tennessine/blob/cb5a2cbcd7ec60d0c7e45aead0c4c4d83218b3aa/packages/stateManager/mod_test.ts#L335-L336

this one is a bit annoying but i'm not sure what to do about it rn

PR #128, opened 2024-11-08
basic frontend tests working

wanderer:
I ported a single test file /tests/AddProductView.test.tsx, which now runs with but doesn't pass
```
 deno test --import-map=./importmap.json -A ./tests/AddProductView.test.tsx --no-check --trace-leaks              
```
It doesn't pass due to a resource leak.  More about this later                        

To do this I did the following
- striped out vitest
- update happydom to the latest version (a fix landed two days ago that fixes globals registration in deno)
- created "tests/global-dom.ts" which setups the dom for the tests
- created mockNavigation.ts and an import map to replace the vitest mocks for next/navigation

## Resource Leak

the resource leak is due to test-utils.ts. When test-utils is imported these lines runs https://github.com/masslbs/Tennessine/blob/a06818e3cf57fe974e3f8fd17adc70fc06687016/packages/frontend/tests/test-utils.tsx#L32-L38
Which setsup some listeners, this should only happy with the `test(''some test", () => {
  // you can setup here
})`

## other things
- the stateManger is exporting a mockClient that is getting used in `test-utils`.  The stateManger should export the stateManger only. We should put it in a tests module
 - mockClient exports stateManger??? 
 - circler imports in client/mod.ts <-> clients/stream.ts


wanderer:
Also
- replaced <next/image> with just . This was causing problems and the benefits of next/image are not used since we are not doing SSR.

PR #151, opened 2024-11-15
store id to bigint

youngbbcoconut:
1. change store id to bigint in frontend 
2. fix false positive test in guest checkout
3. small fix using the new blockchain functions passing arg correctly in getPaymentAddress, etc.

Going to fix the PaymentsArg type in frontend in a different PR


packages/client/guestCheckout_test.ts
```diff
@@ -240,29 +241,6 @@ describe({
           },
         },
       });
-      const stream = guestRelayClient.createEventStream();
```
youngbbcoconut, line 243: the checkout test was a false positive test because of this extra stream declaration

wanderer (approved):

PR #161, opened 2024-11-24
useShopId test

youngbbcoconut:


packages/not-next/src/MassMarketContext.tsx
youngbbcoconut, line 1: shopDetails and currentOrder needs to persist, but with this we can get rid of StoreContext

packages/not-next/src/utils/token.ts
youngbbcoconut, line 1: any util fuction that is related to tokens will live here

Issue #165, opened 2024-11-25
Use useState in useCurrentOrder for storing the current order 


wanderer:
We can turn the use effect callback into the following in oder to capture error and loading states which can be consumed by the UI
```
  const promise = orderFetcher();
  promise.final(() => setIsDone(true)).catch((e) => setError(e))
```

Issue #163, opened 2024-11-25
add shopId dep to baseToken Hook


wanderer:

youngbbcoconut:
https://github.com/masslbs/Tennessine/pull/167/files#diff-a9a94554213b02415f855f385c5dd0f33660721384da389f62e6580e7190accc

PR #167, opened 2024-11-26
use wagmi hook usePublicClient

### LINKED ISSUE #162: use wagmi hook for publicClient
https://wagmi.sh/react/api/hooks/usePublicClient

youngbbcoconut:
fixes #162, #163, #165


packages/not-next/src/hooks/usePublicClientByChainId.ts
```diff
@@ -0,0 +1,31 @@
+import { usePublicClient } from "wagmi";
+import { hardhat, mainnet, sepolia } from "viem/chains";
+
+export default function usePublicClientByChainId(id?: number) {
```
wanderer, line 4: this function should do what it say, get a publicClientByChainID, but it does more then that. But first lets start with getting a publicClientByChainID. To do this we only need to do
```
usePublicClient({
    chainId,
  });

```
I do not think wrapping this in another function make sense since it is fairly minimal. So lets consider what else usePublicClientByChainId is doing.
```
  const chainName = process.env.NEXT_PUBLIC_CHAIN_NAME!;
```
This is getting the chain that the store registry is deployed on. so maybe what we want this function to be called is usePublicClientForRegestry().

packages/not-next/src/hooks/usePublicClientByChainId.ts
```diff
@@ -0,0 +1,31 @@
...
+        chainId = hardhat.id;
+        break;
+      case "sepolia":
+        chainId = sepolia.id;
+        break;
+      case "mainnet":
+        chainId = mainnet.id;
+        break;
+      default:
+        throw new Error(`unhandled chain name ${chainName}`);
```
wanderer, line 22: Don't use a switch statement here. Import using the * to a variable then use importVar[chainname].

packages/not-next/src/hooks/useQuery.ts
```diff
@@ -4,7 +4,7 @@ import { hashMessage } from "@wevm/viem";
 // global cache
 const queryCache = new Map();
 
-export function useQuery<T>(
+export default function useQuery<T>(
```
wanderer, line 7: I was not going to use default for any of the hooks. I was thinking in general that stand alone functions would all use normal exports and class definition should use default exports.

wanderer (changes requested):

packages/stateManager/mod.ts
```diff
@@ -92,6 +95,26 @@ async function storeOrdersByStatus(
   }
   return store.put(status, orders);
 }
+// Synchronize db operations to prevent race conditions.
+const lockAndRelease = <T extends ShopObjectTypes>(
+  store: Store<T>,
+): Store<T> => {
+  const dbMutex = new Mutex();
```
cryptix, line 102: i fear this is not comprehensive. The mutex as to be shared by all stores. There has to be one mutex which governs all gets/puts.

PR #168, opened 2024-12-09
mutex

youngbbcoconut:


packages/stateManager/mod.ts
```diff
@@ -92,6 +95,28 @@ async function storeOrdersByStatus(
...
+        return store.put(key, value);
+      });
+    },
+    get: (key: StoreKey) => {
+      return dbMutex.runExclusive(() => {
+        return store.get(key);
+      });
+    },
+    iterator: () => {
+      return store.iterator();
```
cryptix, line 116: why don't we lock this iterator?

we shouldn't read while writes are happening.

PR #169, opened 2024-12-09
Issues 162, 163, 165

youngbbcoconut:


wanderer:
fixes #162 #163 #165

packages/not-next/src/hooks/useShopId.ts
```diff
@@ -1,9 +1,9 @@
 import { useEffect, useState } from "react";
 import { useSearch } from "@tanstack/react-router";
-// import { ShopId } from "@/types";
+import { ShopId } from "../types.ts";
 
-export default function useShopId() {
-  const [shopId, setShopId] = useState<bigint | null>(null);
+export function useShopId() {
+  const [shopId, setShopId] = useState<ShopId | null>(null);
```
wanderer, line 6: at first I didn't really like using `ShopId` instead of `bigint`. But now I think I see the value in the extra level of indirection as it makes changing underlining type easier if needed

packages/not-next/src/hooks/usePublicClientForRegistry.ts
```diff
@@ -0,0 +1,20 @@
+import { usePublicClient } from "npm:wagmi";
+import * as chains from "npm:wagmi/chains";
+import process from "node:process";
+
+export function usePublicClientForRegistry(id?: number) {
```
wanderer, line 5: this function needs cleaned up. The id should be use a default parameter. Something like this
```suggestion
export function usePublicClient(id = chains[process.env.NEXT_PUBLIC_CHAIN_NAME].id) {
```
Also I would rename it to just `usePublicClient` since it is a general hook.

wanderer (changes requested):
Please add a simple test for usePublicClient hook

packages/not-next/src/hooks/usePublicClientForRegistry.ts
youngbbcoconut: I renamed the file to usePublicClient, but since we are importing usePublicClient from wagmi, I changed the function name itself as useClient. But I will still import it as usePublicClient throughout the app since it's not a default export. 
The test for this hook is in my tests branch so i will merge this PR

youngbbcoconut: The test for this hook is in my tests branch so i will merge this PR

PR #180, opened 2024-12-17
async mutation

youngbbcoconut:


packages/stateManager/mod.ts
```diff
@@ -77,25 +77,70 @@ async function storeOrdersByStatus(
...
+    //Assign the resolve function of this new promise to the resolve variable above so we can resolve it once the current operation is done.
+    this._writing = new Promise<void>((r) => (resolve = r));
+
+    try {
+      await current; // Wait for previous operation
+      return await operation();
+    } finally {
+      resolve!();
+    }
+  }
```
wanderer, line 124:
```suggestion

```
We only need one queue

wanderer (changes requested):

packages/stateManager/mod.ts
```diff
@@ -1244,7 +1292,7 @@ export class StateManager {
       //fromPublicKey in KeyCard manager returns the address from public key as all lowercase.
       await this.keycards.verify(event.signer);
       for (const storeObject of storeObjects) {
-        await storeObject._processEvent(event);
+        await storeObject.processEvent(event);
```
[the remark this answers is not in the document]
youngbbcoconut: oh that's true. so we only need to do it for the individual class functions right?

packages/stateManager/mod.ts
```diff
@@ -77,25 +77,41 @@ async function storeOrdersByStatus(
...
-  const orders = (await store.get(status)) as OrdersByStatus || [];
+  const orders = ((await store.get(status)) as OrdersByStatus) || [];
   orders.push(orderId);
   return store.put(status, orders);
 }
 
 abstract class PublicObjectManager<
   T extends ShopObjectTypes,
 > extends EventEmitter {
+  private _sendingClientRequest: Promise<void | unknown> = Promise.resolve();
```
wanderer, line 88:
```suggestion
  private _sendingQueue: Promise<void | unknown> = Promise.resolve();
```

packages/stateManager/mod.ts
```diff
@@ -213,40 +229,43 @@ class ListingManager extends PublicObjectManager<Listing> {
...
+        assert(
+          Object.values(ListingViewState).includes(update.viewState),
+          `update.viewState ${update.viewState} must be a valid ListingViewState`,
+        );
+        ui.viewState = update.viewState;
+      }
+
+      const requestId = await this.client.updateListing(ui);
+      return eventListenAndResolve<Listing>(requestId, this, "update");
+    });
```
wanderer, line 268: I would keep the validation logic out of the queue, since there is no async work going on in it.
```suggestion
      const ui: schema.IUpdateListing = {
        id: { raw: hexToBytes(update.id!) },
      };

      if (update.price) {
        ui.price = { raw: priceToUint256(update.price, decimals) };
      }
      if (update.metadata) {
        ui.metadata = update.metadata;
      }
      if (update.viewState !== undefined) {
        assert(
          Object.values(ListingViewState).includes(update.viewState),
          `update.viewState ${update.viewState} must be a valid ListingViewState`,
        );
        ui.viewState = update.viewState;
      }

    return this.queueClientRequest(async () => {
      const requestId = await this.client.updateListing(ui);
      return eventListenAndResolve<Listing>(requestId, this, "update");
    });
```

packages/stateManager/mod_test.ts
```diff
@@ -27,7 +27,7 @@ import {
   type Tag,
 } from "./types.ts";
 
-async function setupTestManager() {
+export async function setupTestManager() {
```
wanderer, line 30: Move this to a test utils module if you want to reuse it in the other tests. *_test files should only be for the test runner to consume and should not have exports

wanderer (changes requested):

PR #185, opened 2024-12-19
hook tests

youngbbcoconut:
I ran into a lot of typescript errors with the publicClient. I think the problem was that  jsr:@wevm/viem in the imports map has different return type than npm:viem. And the return type from npm is what matches the return type for wagmi's usePublicClient.
So I am using npm:viem in the individual files for now.


packages/not-next/src/hooks/useClientWithStateManager.ts
```diff
@@ -1,48 +1,44 @@
...
+    return clientStateManager;
+  }, [shopId, relayEndpoint, shopPublicClient, clientStateManager?.shopId]);
+
+  useEffect(() => {
+    if (currentClientStateManager !== clientStateManager) {
+      setClientStateManager(currentClientStateManager);
+    }
+  }, [currentClientStateManager, clientStateManager]);
+
+  const result = useQuery(async () => {}, [keycard, clientStateManager]);
```
youngbbcoconut, line 41: leaving this body blank for now, since I have to rewrite the keycard subscription logic with the useKeycard hook

packages/not-next/src/hooks/useClientWithStateManager_test.tsx
```diff
@@ -0,0 +1,73 @@
...
+      routeTree: rootRoute.addChildren([componentRoute]),
+      history: createMemoryHistory({
+        initialEntries: [shopId ? `/?shopId=${shopId}` : "/"],
+      }),
+    });
+
+    return (
+      <StrictMode>
+        <WagmiProvider config={config}>
+          {/* @ts-expect-error */}
```
youngbbcoconut, line 45: it's expecting `<RouterProvider router={router} />` This is also how we are using it in our `App.tsx`

But if we use that syntax then the router functions don't actually work in testing.

youngbbcoconut:
will revert all the npm:viem and revert deno.lock and delete wagmi install from stateManager/deno.json

packages/not-next/src/hooks/useClientWithStateManager.ts
```diff
@@ -1,48 +1,39 @@
...
+      const csm = new ClientWithStateManager(
+        shopPublicClient,
+        shopId,
+        relayEndpoint,
+      );
+      setClientStateManager(csm);
+    }
+  }, [shopId, relayEndpoint, shopPublicClient]);
+
+  // const result = useQuery(async () => {}, [keycard, clientStateManager]);
```
youngbbcoconut, line 35: commenting out useQuery logic here for now since I have to rewrite the keycard subscription logic with the useKeycard hook

PR #194, opened 2025-01-07
Polyfill

youngbbcoconut:


packages/not-next/deno.json
```diff
@@ -23,10 +23,14 @@
     "@types/react": "npm:@types/react@^18.3.11",
     "@types/react-dom": "npm:@types/react-dom@^18.3.1",
     "@vitejs/plugin-react": "npm:@vitejs/plugin-react@^4.3.3",
+    "autoprefixer": "npm:autoprefixer@^10.4.20",
```
wanderer, line 26: where is this used?

packages/not-next/src/App_test.ts
```diff
@@ -17,9 +15,15 @@ Deno.test("check that we can render the app", async () => {
...
+
+  let error: Error | undefined;
+  try {
+    const { unmount } = render(App({ wagmiConfig: config }));
+    unmount();
+  } catch (e) {
+    error = e as Error;
+  }
+  assertEquals(error, undefined);
+
```
wanderer, line 27: It would be nice if we check that some dom was actully rendered here

packages/not-next/src/components/common/AvatarUpload.tsx
```diff
@@ -0,0 +1,84 @@
...
+          const r = e.target as FileReader;
+          const url = r.result;
+          typeof url == "string" && setLocalImg(url);
+          setImgBlob(blob);
+        };
+
+        reader.readAsDataURL(fileInput.files[0]);
+      }
+    } catch (error) {
+      console.error(error);
```
wanderer, line 39: need to display the error to the user here

packages/not-next/src/hooks/usePublicClient.ts
```diff
@@ -1,13 +1,13 @@
...
 
 export function usePublicClient(
-  id = chains[process.env.NEXT_PUBLIC_CHAIN_NAME as keyof typeof chains].id,
-) {
+  id = chains[import.meta.env?.VITE_CHAIN_NAME as keyof typeof chains]?.id ?? 1,
+): { shopPublicClient: PublicClient } {
   const shopPublicClient = usePC({
     chainId: id,
-  });
+  }) as PublicClient;
```
wanderer, line 10: `as PublicClient;` is probably unessicary. usePC should return the type `PublicClient | Undefined`

packages/not-next/src/utils/token.ts
```diff
@@ -16,3 +17,28 @@ export const getTokenAddress = (symbol: string, chainId: string): Address => {
   }
   return tokenAddress;
 };
+
+export const getTokenInformation = (
```
wanderer, line 21: I would be good to have a test for this function

wanderer (changes requested):

packages/not-next/deno.json
youngbbcoconut: tailwind css doc says it's a peer dependency when using with vite:
https://tailwindcss.com/docs/guides/vite

packages/not-next/src/hooks/usePublicClient_test.tsx
```diff
@@ -26,24 +26,19 @@ Deno.test("usePublicClient", async (t) => {
...
+      const { result } = renderHook(() => usePublicClient(sepolia.id), {
         wrapper,
       });
-      assertEquals(result.current.shopPublicClient?.chain.id, 1);
-    },
-  );
-  await t.step(
-    "should use environment chain when no chainId is provided",
-    () => {
-      Deno.env.set("NEXT_PUBLIC_CHAIN_NAME", "sepolia");
```
youngbbcoconut, line 38: frontend now uses `import.meta.env` 
Can't set vite env var in deno testing

wanderer (approved):

PR #206, opened 2025-01-22
cart and checkout flow issue #192

### LINKED ISSUE #192: Cart Managment Flow

youngbbcoconut:
closes #192


packages/utils/mod.ts
```diff
@@ -146,6 +146,14 @@ export function formatUnitsFromString(price: string, decimal: number) {
   return formatUnits(BigInt(price), decimal);
 }
 
+export function multiplyAndFormatUnits(
```
wanderer, line 149: This function seems to be very specific. It looks like it only used in Cart.tsx. So I think it should live there, at least in not-next.

wanderer (changes requested):

PR #207, opened 2025-01-23
handle duplicate key card error

### LINKED ISSUE #203: subscription requests for different keycards
if the keycard is trying to send a subscription request but the keycard is not enrolled, we should enroll with a new keycard

youngbbcoconut:
closes #203

### DIFF
--- a/packages/not-next/src/hooks/useBaseToken.ts
+++ b/packages/not-next/src/hooks/useBaseToken.ts
@@ -1,18 +1,12 @@
 import { useEffect, useState } from "react";
 
-import { logger } from "@massmarket/utils";
-
 import { useClientWithStateManager } from "./useClientWithStateManager.ts";
 import { usePublicClient } from "./usePublicClient.ts";
 import { getTokenInformation } from "../utils/token.ts";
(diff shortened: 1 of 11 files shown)


packages/not-next/src/hooks/useBaseToken.ts
```diff
@@ -39,19 +33,15 @@ export function useBaseToken() {
     getManifest();
   }, []);
 
-  const { result } = useQuery(async () => {
+  const { result: baseToken } = useQuery(async () => {
```
youngbbcoconut, line 36: return query directly, no need to set it as state first

packages/not-next/src/hooks/useShopId.ts
```diff
@@ -1,18 +1,6 @@
...
-
-  useEffect(() => {
-    const id = search?.shopId;
-    if (id) {
-      setShopId(BigInt(id));
-    }
-  }, [search]);
-
-  return { shopId };
+  return { shopId: search?.shopId ? BigInt(search.shopId) : null };
```
youngbbcoconut, line 5: return shopId directly without setting it as state first

wanderer (approved):

PR #223, opened 2025-02-05
get allowance for erc20

youngbbcoconut:


packages/not-next/src/components/cart/Pay.tsx
```diff
@@ -62,10 +66,28 @@ export default function Pay({
...
         debug("Pending ERC20 contract call approval");
-
+        const allowance = await getAllowance(shopPublicClient!, [
+          paymentArgs[0].payeeAddress,
+          paymentArgs[0].currency,
+        ]);
+        let amount = paymentArgs[0].amount;
+        if (allowance < paymentArgs[0].amount) {
+          amount = paymentArgs[0].amount - allowance;
+        }
```
cryptix, line 76: we should make at least one new variation of our tests where we trigger this code by manually setting a smaller allowance before.

wanderer (approved):

PR #241, opened 2025-02-14
Fix paymentId padding

cryptix:


packages/blockchain/mod.ts
```diff
@@ -111,12 +111,27 @@ export const getPaymentAddress = genericReadContract(
...
+  const result: bigint = await getPaymentIdRaw(...args);
+  const resultAsHex = ("0x" + result.toString(16)) as `0x${string}`;
+  const resultBytes = hexToBytes(resultAsHex);
+  if (resultBytes.length === 32) {
+    return resultBytes;
+  }
+  // if result fits in less then 32bytes we have to fix the padding
+  const paddedResult = new Uint8Array(32);
+  paddedResult.set(resultBytes, paddedResult.length - resultBytes.length);
+  return paddedResult;
```
wanderer, line 132: since we already have viem as a dependency i think it would be better to do
```suggestion
  return pad(numberToBytes(result))
```
and 
```
import { numberToBytes, pad } from 'npm:viem'
```

wanderer (changes requested):

PR #247, opened 2025-02-21
replace fetch and other web apis for happy dom testing

youngbbcoconut:


packages/frontend/src/happyDomSetup.ts
```diff
@@ -1,3 +1,14 @@
...
+const denoHeaders = globalThis.Headers;
+const denoRequest = globalThis.Request;
+const denoAbortController = globalThis.AbortController;
+
 GlobalRegistrator.register();
+
+globalThis.AbortController = denoAbortController;
+globalThis.Headers = denoHeaders;
+globalThis.Request = denoRequest;
+globalThis.fetch = denoFetch;
```
wanderer, line 14: can you see if this works?
```
const oldGlobal = Object.assign({}, globalThis)
GlobalRegistrator.register();
Object.assign(globalThis, oldGlobal)
```
The above saves all of deno's globals the rewrites them back. This might catch other errors in  the futures. But there is also a chance that it breaks happydom. But it it does work I think it would be better.

wanderer (changes requested):

packages/frontend/src/happyDomSetup.ts
youngbbcoconut: hmm I still get the same errors from before  we reassigned AbortController and Headers, like abortSignal is not a function, Cannot read properties of undefined (reading 'headerList')

wanderer: right, so when you do 
```
Object.getOwnPropertyDescriptor(globalThis, "AbortController")
```
you get 

```
{
  value: [class AbortController],
  writable: true,
  enumerable: false,
  configurable: true
}
```
so we need to get the non-enumerable props from globalThis  using Object.getOwnPropertyNames. But now we have other failures

PR #248, opened 2025-02-21
move testFixtures to schema & use hex for shopId in url

youngbbcoconut:


packages/frontend/src/components/cart/Checkout_test.tsx
youngbbcoconut, line 1: just a basic render test for now

wanderer (approved):

PR #261, opened 2025-03-10
WIP - Cbor 

wanderer:


packages/merkle-dag-builder/mod.ts
```diff
@@ -1,313 +1,244 @@
...
+    obj.set(key, value);
+  } else if (
+    typeof obj === "object" && obj !== null &&
+    (typeof key === "number" ||
+      typeof key === "string" ||
+      typeof key === "symbol")
+  ) {
+    Reflect.set(obj, key, value);
+  } else {
+    throw new Error(`Cannot set key ${key} on ${obj}`);
```
cryptix, line 70: this is a bit off topic but i just stumbled over this thought since this function isn't a promise, it might not be obvious to expect this to throw..?

is there something like an "unhandled rejection / exception" linter?

PR #264, opened 2025-03-14
[WIP] move frontend to new stateManager and cbor type definitions

cryptix:


packages/frontend/src/components/merchants/ShopSettings.tsx
```diff
@@ -186,30 +186,39 @@ export default function ShopSettings() {
...
+      newAcceptedCurrencies.set(chainId, allChainAddresses);
+      setAcceptedCurrencies(newAcceptedCurrencies);
     } else {
-      // Remove from accepted currencies
-      const test = acceptedCurrencies.filter((c: Currency) => {
-        return !(c.address.toLowerCase() === address.toLowerCase() &&
-          c.chainId === Number(chainId));
-      });
+      if (allChainAddresses) {
+        //TODO: may have to delete address as lowercase
```
cryptix, line 199: This might be a good case to overload the map with custom get/set that do the lower-casing.

In general it's fine to lower-case the addresses. The casing is actually "just" a checksum: `https://coincodex.com/article/2078/ethereum-address-checksum-explained/`

cryptix: so both versions (capitalized or binary) are functionally equivalent as far as the value of the address is concerned. The checksum also disappears when going from hex to binary.

packages/frontend/src/components/merchants/listings/EditListing.tsx
```diff
@@ -74,55 +79,57 @@ export default function EditProduct() {
...
-      await clientStateManager!.stateManager!.listings.changeInventory(
-        id,
-        units,
-      );
+      await clientStateManager!.stateManager.set("Listing", newItem);
+      // await clientStateManager!.stateManager!.listings.changeInventory(
+      //   id,
+      //   units,
+      // );
+      await clientStateManager!.stateManager!.set("ChangeInventory", {
```
cryptix, line 89: i'm not sure how to fold this best into the api but on the patch/network level we support `increment` for deltas as well as `replace` for new values.

PR #289, opened 2025-03-26
Bubble errors back to stateManager operations

cryptix:
So this was a bit weird.

Afaict stream processing stops if you throw from inside the write function. So if we want to support more then one error happening, we can't use the .catch() style handling.

I therefore ended up removing the old test and am now adding a check if it's a response error, and throw other errors. (so that disconnect for ex. actually defuncts the active stream)

### DIFF
--- a/packages/client/mod.ts
+++ b/packages/client/mod.ts
@@ -60,7 +60,7 @@ export type Patch =
     }
   );
 
-// TODO: maybe there is a better type definitoin for resolve and reject somewhere
+// TODO: maybe there is a better type definition for resolve and reject somewhere
 export type SendingPatchSet = {
   patches: Patch[];
   resolve?: () => void;
@@ -348,7 +348,7 @@ export class RelayClient {
           }
           const writeError = new ClientWriteError(error, signedPatchSet);
           reject?.(writeError);
-          // throw writeError;
+          throw writeError;
           // TODO: abort stream? if not, swallow the error..?
         }
       },
--- a/packages/client/mod.ts
+++ b/packages/client/mod.ts
@@ -341,15 +341,14 @@ export class RelayClient {
         try {
           const result = await this.encodeAndSend(envelope);
           resolve?.(); // could return response but what for..?
-          return result;
+          return result; // where does this result go..?
         } catch (error) {
           if (!(error instanceof Error)) {
             throw new Error(`unknown error: ${error}`);
           }
           const writeError = new ClientWriteError(error, signedPatchSet);
           reject?.(writeError);
-          throw writeError;
-          // TODO: abort stream? if not, swallow the error..?
+          console.error(writeError);
         }
       },
     });
--- a/packages/client/mod.ts
+++ b/packages/client/mod.ts
@@ -52,13 +52,20 @@ export type Patch =
   }
   & (
     | {
-      Op: "add" | "remove" | "replace" | "append";
+      Op: "add" | "remove" | "replace" | "append" | "increment" | "decrement";
       Value: CodecValue;
     }
     | {
-      Op: "remove" | "increment" | "decrement";
+      Op: "remove";
     }
   );
+
+// TODO: maybe there is a better type definitoin for resolve and reject somewhere
+export type SendingPatchSet = {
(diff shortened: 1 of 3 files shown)


packages/client/mod.ts
```diff
@@ -52,14 +52,21 @@ export type Patch =
   }
   & (
     | {
-      Op: "add" | "remove" | "replace" | "append";
+      Op: "add" | "remove" | "replace" | "append" | "increment" | "decrement";
```
cryptix, line 55: a bit unrelated but inc/dec to take a value (the amount you want to change, so you can do += 5 with this for example)

packages/client/mod.ts
```diff
@@ -330,11 +339,17 @@ export class RelayClient {
...
+          const result = await this.encodeAndSend(envelope);
+          resolve?.(); // could return response but what for..?
+          return result;
         } catch (error) {
-          if (error instanceof Error) {
-            throw new ClientWriteError(error, signedPatchSet);
+          if (!(error instanceof Error)) {
+            throw new Error(`unknown error: ${error}`);
           }
+          const writeError = new ClientWriteError(error, signedPatchSet);
```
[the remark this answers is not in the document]
cryptix:
```
[cryptix@crypdesk:~/Mass/Tennessine/packages/stateManager]$ deno test ./mod_test.ts  --no-check -A
running 2 tests from ./mod_test.ts
Database Testings ...
------- output -------
[relayClient] network request subscriptionRequest sent id: 1
[relayClient] network request authRequest sent id: 2
[relayClient] network request response id: 1 received
[relayClient] network request response id: 1 received response
[relayClient] network request response id: 2 received
[relayClient] network request response id: 2 received response
[relayClient] network request challengeSolutionRequest sent id: 3
[relayClient] network request response id: 3 received
[relayClient] network request response id: 3 received response
[relayClient] network request syncStatusRequest id: 1 received
[relayClient] network request subscriptionPushRequest id: 2 received
[relayClient] network request response sent id: 2
----- output end -----
  Make sure stateManager throws an error when a patch is rejected ...
------- output -------
[relayClient] network request patchSetWriteRequest sent id: 5
[relayClient] network request response id: 5 received
----- output end -----
  Make sure stateManager throws an error when a patch is rejected ... ok (42ms)
  add a relay and set a key and retrieve it ...
------- output -------
[relayClient] network request patchSetWriteRequest sent id: 6
[relayClient] network request response id: 6 received
[relayClient] network request response id: 6 received response
----- output end -----
  add a relay and set a key and retrieve it ... ok (4ms)
  Make sure stateManager throws an error when a patch is rejected ...
------- output -------
[relayClient] network request patchSetWriteRequest sent id: 7
[relayClient] network request response id: 7 received
Uncaught error from ./mod_test.ts FAILED
Database Testings ... cancelled (0ms)
Load vector states ... cancelled (0ms)

 ERRORS

./mod_test.ts (uncaught error)
error: (in promise) Error: failed to unmarshal PricingCurrency: cbor: cannot unmarshal UTF-8 text string into Go value of type objects.ChainAddress
          const writeError = new ClientWriteError(error, signedPatchSet);
                             ^
    at Object.write (file:///home/cryptix/Mass/Tennessine/packages/client/mod.ts:349:30)
This error was not caught from a test and caused the test runner to fail on the referenced module.
It most likely originated from a dangling promise, event/timeout handler or top-level code.

 FAILURES

./mod_test.ts (uncaught error)

FAILED | 0 passed (2 steps) | 3 failed (110ms)

error: Test failed
```

packages/stateManager/mod.ts
```diff
@@ -207,26 +212,32 @@ export default class StateManager {
     };
   }
 
-  #sendPatch(state: IStoredState, patch: Patch) {
-    // send patch to peers
+  async #sendPatch(state: IStoredState, patch: Patch) {
+    // send patch to connected peers
+    const allWrites: Promise<void>[] = [];
```
cryptix, line 217: this seemed more stable in the face of having no connections but idk really. Wdyt @wanderer ?

wanderer:
closed; we are writting to the stream directly now.

PR #295, opened 2025-03-27
call initialize then set csm

youngbbcoconut:


packages/frontend/src/ClientWithStateManager.ts
```diff
@@ -5,29 +5,31 @@ import { type IRelayEndpoint, RelayClient } from "@massmarket/client";
...
-    const store = new LevelStore();
-    this.stateManager = new Database(
-      {
-        store,
-        objectId: shopId,
-      },
-    );
+  ) {}
+
+  async initialize() {
```
cryptix, line 18:
```suggestion
  async open() {
```

to not confuse things, personally, i'd prefer to just mirror the name of the wrapped thing

PR #302, opened 2025-04-02
use jollytoad in frontend

youngbbcoconut:


packages/frontend/src/components/merchants/listings/EditListing.tsx
```diff
@@ -154,7 +154,7 @@ export default function EditProduct() {
     e: ChangeEvent<HTMLInputElement>,
     field: string,
   ) {
-    // TODO: why do we need to create a copy here?
+    // We need to create a deep copy of the class and call setListing, or else react will not recognize it as a state change, and will not re-render the component.
```
youngbbcoconut, line 157: If we don't want to do this, we'll have to separate out the components states to something like
const [metadata, setmetadata] = useState(new Metadata()) 
etc.

packages/utils/mod.ts
```diff
@@ -148,11 +148,6 @@ export function priceToUint256(priceString: string, decimals = 18) {
   return numberToBytes(priceInSmallestUnit, { size: 32 });
 }
 
-//Since we are currently storing price as a string, convert string to bigint, then calculate the decimal point.
-export function formatUnitsFromString(price: string, decimal: number) {
```
cryptix, line 152: ah, now i see. Well yes, it's not a string anymore but it's a BigInt and needs to be formatted still.

cryptix (approved):
Nice! I will merge this into my branch manually. Comments mostly left as follow-ups.

PR #310, opened 2025-04-04
frontend tests 

youngbbcoconut:


packages/schema/testFixtures.ts
```diff
@@ -25,4 +25,29 @@ listingVector.get("Snapshots")?.forEach((snapshot) => {
...
+  }
+  for (const [_key, listingMap] of listingsHamt.entries()) {
+    orderListingsMap.set(
+      _key,
+      Listing.fromCBOR(listingMap as Map<string, CodecValue>),
+    );
+  }
+});
+export const allOrders = ordersMap;
+export const allOrderListings = orderListingsMap;
```
youngbbcoconut, line 52: I ended up just using listings and orders from OrderOkay, since the Order IDs in ShopOkay were bigint, the relay was throwing error when trying to send order patch

PR #334, opened 2025-04-11
connect merchant adds client connection

youngbbcoconut:


packages/frontend/src/components/merchants/MerchantConnect.tsx
```diff
@@ -114,11 +117,22 @@ export default function MerchantConnect() {
...
+        role: KeycardRole.MERCHANT,
+        address: keycard.address,
+      });
+
       if (!res.ok) {
         throw new Error("Failed to enroll keycard");
       }
       debug(`Keycard enrolled: ${keycard.privateKey}`);
-      await relayClient.connect();
+      if (!stateManager) {
```
cryptix, line 131: we need to do these asserts before doing any action. Like, we dont want to enroll a keycard and then error out.

PR #341, opened 2025-04-15
update relay once more

cryptix:
also adds back pinned nixpkgs


cryptix:
warning: this update changes the contracts. the production build will not work before the updates are deployed to sepolia and the relay is re-deployed, too.

PR #345, opened 2025-04-16
Update cart quantity + fix pointer hover on top-left logo

lbsal:
The top left element is used as a component to lead a user to the back
to /listings when clicked. We now put cursor: pointer on it, so that a
cursor turns into a pointing hand to signify "hey i am clickable!".

---

This commit changes quantity in ListingDetail.tsx to be of type number
rather than string. We also set defaults of values to be 1 rather than
empty string or 0, as it is a better UX for customers to be able to click
"Add to basket" rather than having to increment first and then click
Add.

The major change concerns what to do when "Add to basket" is clicked for
an item that is already in the cart. Previously the item quantity was fully
replaced with the newly added quantity. A more intuitive behaviour is to
instead add the two quantities such that:

cart(item).quantity = cart(item).quantity + newQuantity

### DIFF
--- a/packages/frontend/src/components/ListingDetail.tsx
+++ b/packages/frontend/src/components/ListingDetail.tsx
@@ -20,7 +20,7 @@ import SuccessToast from "./common/SuccessToast.tsx";
 import { useCurrentOrder } from "../hooks/useCurrentOrder.ts";
 
 const namespace = "frontend:listing-detail";
-const describe = (subNamespace) =>
+const describe = (subNamespace: string) =>
   logger(`${namespace}:${subNamespace}`, "warn");
 const warn = logger(namespace, "warn");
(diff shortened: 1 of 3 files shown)


packages/frontend/src/components/ListingDetail.tsx
```diff
@@ -20,6 +20,8 @@ import SuccessToast from "./common/SuccessToast.tsx";
 import { useCurrentOrder } from "../hooks/useCurrentOrder.ts";
 
 const namespace = "frontend:listing-detail";
+const describe = (subNamespace) =>
```
wanderer, line 23:
```suggestion
const describe = (subNamespace: string) =>
```

cryptix (approved):

cryptix:
Merging. I've been using this branch a bunch.

PR #346, opened 2025-04-17
fix(nav/cart): basket icon is correctly reset after successful checkout; entire menu row is now clickable

lbsal:
In the hook `useCurrentOrder`, we "reset" `currentOrder` for the following states:
* OrderState.Canceled
* OrderState.Paid.

This resetting is done by setting currentOrder to null, causing this bug.

The bug raises the question of using a sentinel value, instead of null, to represent reset orders where we would do e.g.

currentOrder = SENTINEL_ORDER

where SENTINEL_ORDER is a fake order set as a constant and checked against in addition to simple null checks.

---
Also:

[fix(nav/menu): make entire menu row clickable, not just text](https://github.com/masslbs/Tennessine/pull/346/commits/1bfa626db307240a421139057ea2706fd078a633)

### DIFF
--- a/packages/frontend/src/components/Navigation.tsx
+++ b/packages/frontend/src/components/Navigation.tsx
@@ -175,31 +175,31 @@ function Navigation() {
           key={i}
           onClick={() => setMenuOpen(false)}
         >
-          <div className="flex gap-3 items-center">
-            <img
-              src={`/icons/${opt.img}`}
-              width={20}
-              height={20}
-              alt="menu-item"
-              className="w-5 h-5"
-            />
-            <Link
-              to={opt.href!}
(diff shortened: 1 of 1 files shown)


cryptix:
> No idea what the failing test is though

Yea, this was a thing before and wasn't caused by the PR, I'm sure.


There is a race somewhere in the closing code of store or stateManager, I think. I couldn't make sense of the values before, so I added the "this should not be the old value" assert, but it actually is the old value still for some reason...

cryptix:
rebased on #345

wanderer:
lgtm

wanderer:
also i see the issue with the above fail test, will push a fix in the morning

Issue #366, opened 2025-04-23
Beef up testing of checkout flow corner case testing


cryptix:
user testing showed the app tried to clear the basket when proceeding to checkout with insufficient stock.

additionally should show out-of-stock error code better before trying to clone the order, which is why we are seeing "error clearing cart" on the checkout button.

cryptix:
I'm not sure if I have the full picture but there is a spectrum of things here. Let's start with what we can do without bubbling up changes to network-schema / the relay.

1. We should be able to just show "out of stock" error
2. We should make sure when we clear & clone the cart, and not do so in this case. (We might be just hiding the out-of-stock error because we immediately try to re-create it, running into the same issue with the new cart. This hypothesis needs to be tested, tho.)

Now, this error alone doesn't tell the user much. Since we don't include the ListingID in the error code they still need to guess but it's less opague. (For historical reasons the ID isn't there right now. We used to fire these when adding an item. We are now checking quantities at checkout for all items together). So, I think next should be exposing that in some structural manor. 

3. Add ListingID to OutOfStock errors (probably should do this for NotFound, too just for consistency). Then we could show the affected Listing in the toaster, for example.
4. I'd also argue we should add more 'non-happy path tests', especially around these "do we need to clone the cart and start over" cases.

cryptix:
> `MIN(stock_remaining, chosenQuantity)`

`stock_remaining` isn't directly accessible to a customer. The exact inventory is private data to the merchant. We do have `StockStatuses` on items but these are not tied used or set by the frontend yet.

wanderer:
@cryptix the cart is just in the front-end now and doesn't exist on the relay correct? We only have orders on the relay

cryptix:
Yes. Error code / naming is outdated

cryptix:
I'm relatively certain the problem lies here:

https://github.com/masslbs/Tennessine/blob/c564b4f0d4fa99e29a6eb6a3c765986a16156ebd/packages/frontend/src/components/cart/Cart.tsx#L110-L153

`handleCheckout()` checks the error but then `clearCart()` fails and hides it.

PR #370, opened 2025-04-23
WIP: use additional_info.object_id from error

cryptix:
updates #366

### DIFF
--- a/packages/frontend/src/components/cart/Cart.tsx
+++ b/packages/frontend/src/components/cart/Cart.tsx
@@ -19,6 +19,7 @@ import { useBaseToken } from "../../hooks/useBaseToken.ts";
 import { useCurrentOrder } from "../../hooks/useCurrentOrder.ts";
 import { useStateManager } from "../../hooks/useStateManager.ts";
 import { multiplyAndFormatUnits } from "../../utils/helper.ts";
+import { RelayResponseError } from "../../../../client/mod.ts";
 
 const namespace = "frontend:Cart";
 const debug = logger(namespace);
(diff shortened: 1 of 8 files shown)


packages/frontend/src/components/cart/CheckoutFlow.tsx
```diff
@@ -133,20 +133,15 @@ export default function CheckoutFlow() {
...
-        await stateManager.set(
-          ["Orders", currentOrder.ID, "State"],
-          OrderState.Committed,
-        );
-        debug(`Order ID: ${currentOrder.ID} committed`);
-      }
-      setStep(CheckoutStep.shippingDetails);
-    } catch (error: unknown) {
-      logerr("Error during checkout", error);
-      throw error;
```
cryptix, line 148: we don't need to try-catch this block, since we catch and log it on the other call site already.

packages/frontend/src/components/cart/Checkout_test.tsx
```diff
@@ -155,14 +157,19 @@ Deno.test("Check that we can render the checkout screen", {
...
-      expect(shippingScreen).toBeTruthy();
+      const errorToaster = screen.getByTestId("error-message");
+      expect(errorToaster).toBeTruthy();
+      expect(errorToaster.textContent).toContain("Item 23 is out of stock.");
     });
   });
 
+  // TODO:
+  // - evaluate the error toaster (referencing the out of stock item)
+  // -
```
cryptix, line 170: this tests needs a little love, removing the item and then trying again

packages/frontend/src/components/cart/Cart.tsx
```diff
@@ -110,16 +113,15 @@ export default function Cart({
...
-        )
+        // TODO: This is a big turd but i couldn't make it recdognize RelayResponseError for the life of me.
+        error.originalError.code === 9 && error.originalError.additionalInfo
       ) {
-        setErrorMsg("Not enough stock. Cart cleared.");
-        await clearCart();
+        const objectId = error.originalError.additionalInfo.object_id;
+        console.log("Object ID:", objectId);
+        setErrorMsg(`Item ${objectId} is out of stock.`);
+        // await clearCart();
```
cryptix, line 124: this is just stapled together, too. should make a nicer toaster message using the item title, etc.

cryptix:
superseeded by #398

PR #398, opened 2025-04-24
Use error.additionalInfo for outOfStock errors

youngbbcoconut:


cryptix (approved):
Patched out one minor wrinkle. LGTM now!

PR #424, opened 2025-04-29
only warn about unavailable subscription handlers

cryptix:
since we cancel subscriptions, we can't assume the controller _must_
be there when receiving push requests.

https://github.com/masslbs/Tennessine/blob/8980cc01c47452c93dd14c616659dd29471467e0/packages/client/mod.ts#L341-L344

A push might still be in transit, which needs to be dropped without breaking processing/operations.


wanderer:
The trouble is if you have a path like `/some/0/thing/else`  you don't know if `0` is an index into a array or a key on a map.

cryptix:
Ah.. I see! Okay, let's punt on this then until we have locale state verification.

cryptix:
i'd kinda argue that we should warn and skip unhandled patches instead of breaking on them but might be separate issue.

PR #426, opened 2025-04-29
Vite script

wanderer:
this uses scripts instead of the builtin vite command to run a dev server and do production builds.


packages/frontend/src/release-info.json
```diff
@@ -1,6 +1,6 @@
 {
-  "version": "v0.1-dev",
-  "commit": "unset",
+  "version": "545d9a83",
+  "commit": "545d9a834019b0fbb7c0ea594320f9b65b4480e4",
   "tag": null,
-  "timestamp": "unset"
+  "timestamp": "2025-04-29T20:05:42.000Z"
```
cryptix, line 5: let's not commit these. I think i have a better idea of doing this once we have #418 merged (basically append to `.env.production` as the pre-step to `build`)

PR #433, opened 2025-05-01
moved webstream pollyfil to the frontend

wanderer:


wanderer:
I moved the pollyfill to here https://github.com/masslbs/Tennessine/pull/433/files#diff-b84f774cee1ab7736e8199faad69a3d59dab96adde7c466b81acb0820297662eR1

PR #479, opened 2025-05-09
Add more info to success toast + improve test flake

lbsal:
Changed the success toast displayed on ListingDetail to include more information about the quantity that was just added. Then used this information to improve the ListingDetail test by using a findByText query, which waits for the appearance of its input for a certain amount of time.

The hope is that waiting for the success toast to change gives the stateManager logic enough time to run and actually finish updating the quantity internal, making this particular test flake less often in CI.


cryptix:
Also noting that all of these show a "There was an error updating cart". Maybe we should assert first that there was no error? This might give us better visibility on the actual problem/race.

PR #489, opened 2025-05-13
user events should be wrapped around act function

youngbbcoconut:


wanderer:
so the thing im not sure about is why `act` fixes this. Any insight to why?

Issue #499, opened 2025-05-15
Race condition in stateManager


lbsal:
Relates to https://github.com/masslbs/Tennessine/issues/420, and specifically https://github.com/masslbs/Tennessine/issues/420#issuecomment-2850535011

**The problem**: When running `ListingDetail_test.tsx` the test intermittently fails in the step that exercises the function `hooks/useCurrentOrder:cancelAndRecreateOrder()`.

The reason for the test failure is that the newly created order cannot be retrieved i.e. the new order is not present in the database (i.e. `stateManager`). 

As this only happens intermittently, and especially when the system is placed under load (see https://github.com/masslbs/Tennessine/issues/420#issuecomment-2850535011) this has the look of a race condition. After extensive investigation, I can confirm that is so: a failing test retries a few times before finally erroring out, and during that timeout period the new order becomes available in the database (albeit too late). 

As it seems to be a race condition, and since I could replicate it, I decided to look further into it because it seems like there's a root cause needing to be addressed.

The relevant sequence of events looks something like this (simplified):

* (`ListingDetail.tsx`) cancel and recreate the current order
* (`useCurrentOrder.tsx`) perform operations on state manager to mark the previous current order as canceled
* (`useCurrentOrder.tsx`) copy over the items of the canceled order into a new order
* (`useCurrentOrder.tsx`) **store the new order in the database**: `stateManager.set("Orders", <newOrderID>,
  <newOrderObject>`
  * `stateManager.set` in turn calls `this.graph.set(state.root, path, value)`, for some value
    of state.root
* (`ListingDetail.tsx`) get **the new order from the database**: `stateManager.get("Orders", <newOrderID>)`
    * `stateManager.get` in turn calls `this.graph.get(state.root, path)`, for some value of
      state.root (! this becomes important later)

What I have looked into:

* whether the race condition relates to an incoming patch from the relay being processed too slowly
* if there is a bug inside `stateManager` wrt `set()` & `get()`
* if there is a bug inside `merkle-dag-builder`
* if there was some misuse along the sequence of events `ListingDetail` -> `useCurrentOrder` ->
  `stateManager` and back
* whether there are two different stateManagers being operated on by the customer (there was not)

Observations:

[...]

PR #501, opened 2025-05-15
State manager promise strictness

wanderer:
This is in response to https://github.com/masslbs/Tennessine/issues/499
There is a possible race condition in the stateManger. Using  `state.root = await X` causes a race condition,  `state.root` must always be set atomically. Lets consider why. 

Lets say the stateManger recieves a patchset from the relay. While applying the patch's it sets the state to  `state.root<v1> = await updatingTheStateFunction`. 

Now lets say stateManger.set is called. at this point `updatingTheStateFunction`  may have not resolved and state.root is now stale as stateManger.set is operating on `state.root<v1>`. When `stateManger.set` finishes its update it sets a new state root: `state.root<v2>`. Lastly `updatingTheStateFunction` finishes and `state.root = await updatingTheStateFunction` sets the state.root to `state.root<v3>` = Result(updatingTheStateFunction(`state.root<v1>`)) which reverts state.root<v2> produced by stateManger.set

### DIFF
--- a/packages/stateManager/mod.ts
+++ b/packages/stateManager/mod.ts
@@ -100,25 +100,25 @@ export default class StateManager {
           // TODO validate the Operation's value if any
           // console.log("Applying patch:", patch);
           if (patch.Op === "add") {
-            state.root = await this.graph.add(
+            state.root = this.graph.add(
               state.root,
               patch.Path,
               patch.Value,
             );
           } else if (patch.Op === "replace") {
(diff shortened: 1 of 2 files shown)


cryptix:
i'm getting none of the failures @lbsal reported but i'm seeing this one every other run: https://gist.github.com/cryptix/9fcb18f74de2f8414673f7e7e09c311e#file-fail1-txt

PR #514, opened 2025-05-16
redirect root route "/" to mass.market

youngbbcoconut:


aeluteia:
> 
> i think in the medium term it would better to throw up a generic view that informs the visitor that there is something wrong "couldn't find shop id!" and provides some way to go to a shop (e.g. an input element that accepts a shopid || full shop url)
> 
> for this particular very short-term use case (the mass market shop is the only possible merchant), i wonder if it's not better to ensure that we always redirect to the current shop. the reason being that if you expect to go to `<shop site>` but somehow end up on `<main site>` that's a jarring transition and reversal of expectations.
> 

Good point, on both. Agree that right now it would be better to redirect straight to our demo/test shop. 

Medium(/long) term there are several elements that it might be good to have on that page - including a way to find a shop by entering its id. This is def a discussion that we should add to the list for when we plan the next chapter.

youngbbcoconut:
to follow up with: https://github.com/masslbs/Tennessine/issues/517

PR #519, opened 2025-05-19
Keycard refactor

youngbbcoconut:


packages/frontend/src/hooks/useKeycard.ts
```diff
@@ -1,27 +1,80 @@
...
-export function useKeycard(
-  keycard?: {
-    privateKey: `0x${string}`;
-    role: KeycardRole;
-    address: `0x${string}`;
-  },
-) {
-  if (!keycard) {
+    if (!shopId) {
+      assert(shopId, "shopId is required");
```
wanderer, line 35: assert throw an error. So this is redundent

packages/frontend/src/hooks/useKeycard.ts
```diff
@@ -1,27 +1,80 @@
...
 
-export function useKeycard(
-  keycard?: {
-    privateKey: `0x${string}`;
-    role: KeycardRole;
-    address: `0x${string}`;
-  },
-) {
-  if (!keycard) {
+    if (!shopId) {
```
wanderer, line 34: we have 3 if statements that do the same thing, we can combine them

packages/frontend/src/wagmi.ts
```diff
@@ -1,8 +1,14 @@
-import { getDefaultConfig } from "@rainbow-me/rainbowkit";
-import { hardhat, mainnet, optimism, sepolia } from "wagmi/chains";
+import { connectorsForWallets, getDefaultConfig } from "@rainbow-me/rainbowkit";
+import {
+  metaMaskWallet,
+  walletConnectWallet,
+} from "@rainbow-me/rainbowkit/wallets";
+import { rainbowkitBurnerWallet } from "burner-connector";
```
wanderer, line 6: I was think that we don't want the burner wallet to show up in rainbowkit, this looks like it is doing the opposite. The reason, being is that it might confuse uses to have a new wallet created for them and already connected. 

The question is, if we do connect the burner wallet automtically will it still show up in places where it shouldn't like the check dialog?

packages/frontend/src/hooks/useKeycard_test.tsx
```diff
@@ -0,0 +1,80 @@
...
+  }, [wallet, shopId, relayEndpoint]);
+
+  return <p data-testid="keycard-role">{keycard?.role}</p>;
+};
+
+Deno.test("useKeycard", {
+  sanitizeResources: false,
+  sanitizeOps: false,
+}, async (t) => {
+  await t.step(
```
wanderer, line 40: I think that all the steps here could be top level `Deno.tests` since they all look like independ tests instead of multiple steps in one tests. (i know we have been using `steps` like this  though)

wanderer (changes requested):

wanderer:
The big thing that is missing here is that multiple components could try to enroll a keycard at the same time. Lets write a test for this first, where to components try to use useKeycard concurrently.  The tests should obserive two different keycards being set. The tests should fail if the components render more then once with a different keycard and pass if they don't. Lets add this test, make sure it fails, then add server synchronization logic in the next step.

packages/frontend/src/wagmi.ts
youngbbcoconut: if user doesn't already have a wallet connected, then this burner wallet will automatically be shown as connected in the rainbowkit modals. Based on burner-connector docs there isn't a straight forward way to prevent rainbow kit from automatically connecting to the burner wallet if no other wallet is connected. I think it might be much more straightforward and more code readable to just use `createWalletClient` instead of this burner connector  to not confuse the user with the rainbowkit modals.

youngbbcoconut:
> The big thing that is missing here is that multiple components could try to enroll a keycard at the same time. Lets write a test for this first, where to components try to use useKeycard concurrently. The tests should obserive two different keycards being set. The tests should fail if the components render more then once with a different keycard and pass if they don't. Lets add this test, make sure it fails, then add [async server synchronization](https://tanstack.com/query/latest/docs/framework/react/overview) logic in the next step.

Not sure I understand - since enroll keycard only happens when we explicitly call the addKeycard method. Not when useKeycard is called.

packages/frontend/src/wagmi.ts
youngbbcoconut: use burner() in connect function

wanderer:
Had a call with @youngbbcoconut we decided
- to leave burner wallet as is for now
- useKeycard can directly `return useQuery`  from tanstack 
- need to check to make sure that tanstack will not invalidate the cache (we only need to enroll a keycard once every (per shop))

packages/frontend/src/hooks/useKeycard.ts
youngbbcoconut, line 1: There is a caveat to enrolling keycard inside useQuery when the hook is called instead of being able to explicitly call enroll. If the hook is being called concurrently with two different roles passed as its arg it may enroll the guest keycard first, then the merchant keycard. But the initial guest keycard enrolled would be useless.

i.e. on` /merchant-connect` the merchant connect component would call `useKeycard("merchant")` but the navigation component which is rendered in every screen  would call `useKeycard("guest")`

packages/frontend/src/hooks/useKeycard.ts
```diff
@@ -1,27 +1,80 @@
...
-  return useLocalStorage(keyCardID, keycard);
+  const { data: wallet } = useWalletClient();
+  const { relayEndpoint } = useRelayEndpoint();
+  const [keycard, setKeycard] = useLocalStorage<Keycard>(
+    "keycard" + shopId,
+    null,
+  );
+
+  return useQuery({
+    queryKey: ["keycard", shopId?.toString(), relayEndpoint, wallet, role],
```
wanderer, line 33: does `relayEndpoint` and `wallet` need to be part of the queryKey?

packages/frontend/src/hooks/useKeycard.ts
```diff
@@ -1,27 +1,80 @@
...
+    queryFn: async () => {
+      // 1. Generate KC
+      // 2. Enroll KC
+      // 3. Save to localStorage
+
+      // This is for instances when the user has a guest keycard and wants to log in as a merchant.
+      const guestToMerchant = role === "merchant" && keycard?.role === "guest";
+      if (keycard && !guestToMerchant) {
+        return keycard;
+      }
```
wanderer, line 43: when would this come into play? is it possible to remove it?

packages/frontend/src/hooks/useKeycard.ts
```diff
@@ -1,27 +1,80 @@
...
+      if (!shopId || !relayEndpoint || !wallet) {
```
wanderer, line 44: Do we need this if statement? Why or why not?

packages/frontend/src/hooks/useKeycard.ts
```diff
@@ -1,27 +1,80 @@
...
+    queryKey: ["keycard", shopId?.toString(), relayEndpoint, wallet, role],
+    queryFn: async () => {
+      // 1. Generate KC
+      // 2. Enroll KC
+      // 3. Save to localStorage
+
+      // This is for instances when the user has a guest keycard and wants to log in as a merchant.
+      const guestToMerchant = role === "merchant" && keycard?.role === "guest";
+      if (keycard && !guestToMerchant) {
+        return keycard;
```
wanderer, line 42: im not sure this works. A merchant can login as a guest though

wanderer: maybe you can use https://tanstack.com/query/latest/docs/reference/QueryClient/#queryclientsetquerydata to set the guest keycard when the merchant is enrolled.

wanderer:
i think we might need set the garbage collection and staleTime

packages/frontend/src/hooks/useKeycard.ts
```diff
@@ -1,27 +1,80 @@
...
+const logger = getLogger(["mass-market", "frontend", "useKeycard"]);
+
+export function useKeycard(role: KeycardRole = "guest") {
   const { shopId } = useShopId();
-  const keyCardID = "keycard" + shopId;
-  // console.log({shopId, keyCardID});
-  return useLocalStorage(keyCardID, keycard);
+  const { data: wallet } = useWalletClient();
+  const { relayEndpoint } = useRelayEndpoint();
+  const [keycard, setKeycard] = useLocalStorage<Keycard>(
```
wanderer, line 27: Do we need this hook?

packages/frontend/src/hooks/useKeycard.ts
```diff
@@ -1,27 +1,80 @@
...
+import { getLogger } from "@logtape/logtape";
+import { useWalletClient } from "wagmi";
+import { useQuery } from "@tanstack/react-query";
+
+import { RelayClient } from "@massmarket/client";
+import { getWindowLocation } from "@massmarket/utils";
 
 import { useShopId } from "./useShopId.ts";
 import { KeycardRole } from "../types.ts";
+import { useRelayEndpoint } from "./useRelayEndpoint.ts";
```
wanderer, line 13: this is going to cause problem. We are also going to need to change this over to tanstack's useQuery. For now maybe do something awful and just create a function in here or seperate file that return the releyEndpoint infromation a promise. We then will call and await it in the body of the below useQuery.

packages/frontend/src/hooks/useKeycard.ts
youngbbcoconut: because wallet and relayEndpoint are not loaded on the first mount, we have to wait for them to actually have values

packages/frontend/src/hooks/useKeycard.ts
youngbbcoconut: we need to wait for all these to have values, it takes a couple renders for these values to be available.

packages/frontend/src/hooks/useKeycard.ts
```diff
@@ -1,27 +1,71 @@
...
-  return useLocalStorage(keyCardID, keycard);
+  const { data: wallet } = useWalletClient();
+  const account = useAccount();
+  const { relayEndpoint } = useRelayEndpoint();
+  const queryClient = useQueryClient();
+
+  return useQuery({
+    queryKey: [
+      "keycard",
+      role,
```
wanderer, line 26: The queryKey need to the account in it. ideally `account.address` since the keycard is tied to the account

packages/frontend/src/hooks/useKeycard.ts
```diff
@@ -1,27 +1,71 @@
...
+
+      const kc = {
+        privateKey,
+        role,
+        address: account!.address,
+      };
+      if (role === "merchant") {
+        // Return this keycard for all keycard queries.
+        queryClient.setQueriesData(
+          { queryKey: ["keycard"] },
```
wanderer, line 62: I don't think this will work https://tanstack.com/query/latest/docs/framework/react/guides/query-keys#if-your-query-function-depends-on-a-variable-include-it-in-your-query-key

You might need something like     queryKey: [ "keycard",    role, account.address ]

wanderer (changes requested):

packages/frontend/src/hooks/useKeycard.ts
```diff
@@ -1,27 +1,69 @@
...
-      address: account.address,
-    };
-  }
+export function useKeycard(role: KeycardRole = "guest") {
   const { shopId } = useShopId();
-  const keyCardID = "keycard" + shopId;
-  // console.log({shopId, keyCardID});
-  return useLocalStorage(keyCardID, keycard);
+  const { data: wallet } = useWalletClient();
+  const account = useAccount();
```
wanderer, line 18: we just care about the address
```suggestion
  const { address } = useAccount();
```

packages/frontend/src/App.tsx
```diff
@@ -1,12 +1,26 @@
...
 import { MassMarketProvider } from "./MassMarketContext.ts";
 import { config } from "./wagmi.ts";
 import { routeTree } from "./routeTree.gen.ts";
 
-const queryClient = new QueryClient();
+export const queryClient = new QueryClient({
+  defaultOptions: {
+    queries: {
+      gcTime: Infinity,
+      staleTime: Infinity,
```
wanderer, line 15: I think this is going to set the gcTime and staleTime for all queryClient globally in the app. Lets move these options to the useQuery in the keycard hook

wanderer (changes requested):

packages/frontend/src/hooks/useKeycard.ts
```diff
@@ -1,27 +1,69 @@
...
+      };
+      // Return this keycard for all guest keycard queries. This is needed for merchant enrolls.
+      queryClient.setQueriesData(
+        { queryKey: ["keycard", account?.address, "guest"] },
+        kc,
+      );
+
+      return kc;
+    },
+    enabled: !!shopId && !!wallet && !!relayEndpoint && !!account,
```
wanderer, line 67: this would make sense the first argument, since when reading this code the first thing you want to think about is the condition under which the following code will run. think of it like an `if` statement. the truthy condition come before the body.

packages/frontend/src/hooks/useKeycard_test.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { expect } from "@std/expect";
+import { random256BigInt } from "@massmarket/utils";
+
+import { createRouterWrapper } from "../testutils/mod.tsx";
+import { useKeycard } from "./useKeycard.ts";
+import { KeycardRole } from "../types.ts";
+
+Deno.test("Should enroll guest keycard", {
+  sanitizeResources: false,
+  sanitizeOps: false,
```
wanderer, line 18: don't repeat these, create options variable.

packages/frontend/src/hooks/useKeycard.ts
```diff
@@ -1,27 +1,69 @@
...
+      const res = await relayClient.enrollKeycard(
+        wallet,
+        account,
+        role === "guest",
+        getWindowLocation(),
+      );
+
+      if (!res.ok) {
+        throw new Error(`Failed to enroll keycard: ${res}`);
+      }
```
wanderer, line 51: Also would be nice to have a test that fail to enroll a keycard.

packages/frontend/src/hooks/useKeycard.ts
```diff
@@ -1,27 +1,69 @@
...
-  if (!keycard) {
-    const privateKey = generatePrivateKey();
-    const account = privateKeyToAccount(privateKey);
-    keycard = {
-      privateKey,
-      role: KeycardRole.NEW_GUEST,
-      address: account.address,
-    };
-  }
+export function useKeycard(role: KeycardRole = "guest") {
```
wanderer, line 15: It would be nice to document this function, check this guide for more clarity https://jsr.io/docs/writing-docs

packages/frontend/src/hooks/useKeycard.ts
```diff
@@ -1,27 +1,83 @@
...
+        role,
+        address,
+      };
+      // Return this keycard for all guest keycard queries.
+      // This is needed for merchant enrolls, since we don't want enrolled merchant keycards to be overwritten by guest keycards.
+      // setQueriesData also ensures that any component using the query will re-render with the new keycard.
+      queryClient.setQueriesData(
+        { queryKey: ["keycard", address, "guest"] },
+        kc,
+      );
```
wanderer, line 75: we only want to set this when enrolling a merchant. Also in the comment its say " since we don't want enrolled merchant keycards to be overwritten by guest keycards." this is not quite what we are doing, we are overwritting the the guest keycard with the merchants 
```suggestion

if (merchant) {
      queryClient.setQueriesData(
        { queryKey: ["keycard", address, "guest"] },
        kc,
      );
    }
```

packages/frontend/src/hooks/useKeycard.ts
```diff
@@ -1,27 +1,83 @@
...
+  const queryClient = useQueryClient();
+
+  return useQuery({
+    // queryFn will not execute till these variables are defined.
+    enabled: !!shopId && !!wallet && !!relayEndpoint && !!address,
+    queryKey: [
+      "keycard",
+      address,
+      role,
+    ],
```
wanderer, line 34: we are using this array twice, if we move it to a variable `const  queryKey = [... ]` it will be easier in the future to update the key if need and will reduce errors since we only have to update it in one place.

wanderer, line 34: oh wait! we need what happen to `shopid` ? we defaintly need that to be in the key. A merchant keycard on one shop will not work on a different shop

packages/frontend/src/hooks/useKeycard.ts
```diff
@@ -1,27 +1,83 @@
...
-    const privateKey = generatePrivateKey();
-    const account = privateKeyToAccount(privateKey);
-    keycard = {
-      privateKey,
-      role: KeycardRole.NEW_GUEST,
-      address: account.address,
-    };
-  }
+/**
+ * This hook returns a keycard for the browser session.
```
wanderer, line 16: i would be more explicit here, the hook return a keycard for a shop given a role which currently is either a "merchant" or "guest" from tanstack's cache.

packages/frontend/src/hooks/useKeycard.ts
youngbbcoconut: the second queryKey array is slightly different because it has to be explicitly "guest" and not _role_

Issue #500, opened 2025-05-15
Use markdown for product description


lbsal:
As a merchant, being able to format product descriptions using rich text is more appealing and expressive than using only plaintext. Markdown is a widely accepted format for accomplishing this.

Note: for the chosen solution there needs to be extensive user generated content sanitization, similar to https://github.com/microcosm-cc/bluemonday's capabilities.

wanderer:
I think this a fun little improvement. Something to keep in minds is that we have yet since we haven't started spec'ing out store customization. Originally I was thinking we could have a geocities like functionality where you could apply your own css & ect. I don't foresee any problems with adding markdown though.

PR #525, opened 2025-05-19
fix(ListingDetail): display description w/ newlines intact

### LINKED ISSUE #503: Split text description input by newline
Split text description input by newlines, then iterate over contents and output as an array inside the product listing component's description

related to #500 but a lot simpler and addresses the use case @aeluteia wrote about in chat

lbsal:
closes #503

note: This patch assumes that `<listing>.Metadata.Description` will always be a string

Tangentially related to #500

### DIFF
--- a/packages/frontend/src/components/ListingDetail.tsx
+++ b/packages/frontend/src/components/ListingDetail.tsx
@@ -140,6 +140,10 @@ export default function ListingDetail() {
     }
   }
 
+  function splitTextByNewlines(input: string) {
+    return input.split("\n");
+  }
+
   return (
     <main
       className="bg-gray-100 md:flex justify-center"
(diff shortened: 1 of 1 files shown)


packages/frontend/src/components/ListingDetail.tsx
```diff
@@ -221,7 +225,15 @@ export default function ListingDetail() {
           <section className="flex gap-4 flex-col bg-white mt-5 md:mt-0 rounded-md md:w-2/5 p-4">
             <div>
               <h3 className=" ">Description</h3>
-              <p data-testid="description">{listing.Metadata.Description}</p>
+              <section data-testid="description">
+                {splitTextByNewlines(listing.Metadata.Description).map(
+                  (line: string, index: number) => (
+                    <p key={`description-${index}`} className="min-h-[1ch]">
```
cryptix, line 231: pr525-2025-05-19
just a minor quibble / react learning Q: is the `key` property special somehow?

cryptix (approved):
tried it and works!

PR #533, opened 2025-05-21
fix grey out for page when nav modal is open

### LINKED ISSUE #392: When basket modal has been triggered, the screen is still scrollable if hovering over navigation
As title. Procedure: click basket icon, the 60% opacity overlay covers all of the screen except the top navigation and scrolling doesn't work (as intended). However, when hovering over the nav, scrolling is still possible. 

Noting it down here for now as a low priority piece of jank that needs fixing. Found during PR #376 but not because of it.

youngbbcoconut:
**1.** truncates super long prices 
image

**2.** grey overlay covers the entirety of the screen when one of the nav modals are open. (closes #392)

image

image

image

### DIFF
--- a/packages/frontend/src/components/CustomerViewListings.tsx
+++ b/packages/frontend/src/components/CustomerViewListings.tsx
@@ -74,7 +74,7 @@ export default function CustomerViewProducts({
                   className="w-5 h-5"
                   data-testid="coin-icon"
                 />
-                <p data-testid="product-price">
+                <p data-testid="product-price" className="truncate">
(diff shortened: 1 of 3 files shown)


youngbbcoconut:
> i can still somewhat reproduce this on my machine
> 
> 2025-05-22-151616_743x348_scrot
> 
> what happens: i can scroll the background / grayed out page while the basket is open by e.g. scrolling while hovering over the navigation element

ah, okay. This is a separate styling issue from the gray overlay not filling the whole screen though so I will make a new issue out of that one. @lbsal

PR #542, opened 2025-05-23
This limits setting the ENS contenthash on mainnet to mainnet releases

wanderer:


cryptix:
Works if you run on the correct branch! :)

https://github.com/masslbs/Tennessine/actions/runs/15210119533

PR #548, opened 2025-05-23
useKeyCard - Using tanstack query for global async state mangement

wanderer:
This is a draft to use tanstack to handle the global keycard state. More generally this draft is testing the waters on whether we should use tanstack for global state async management. Part of the rational to use tanstack is that we where already using it because of wagmi, so this does not introduce a new dependency.


wanderer:
typechecking should now be working on this banch. But we have a few other issues. A few components expect to directly manipulate the keycard namely 
- merchant-connect
- create-shop

I think this issues can be worked around. But I am not sure yet.

PR #553, opened 2025-05-26
DRAFT try to use donenv to get the .env file

wanderer:
This was an attempted to decouple the .env from the frontend dir. This does that, but with the downside that .env resolution is now relative instead of fixed. For testing, this works. But for building this might introduce a foot-gun since you need to build in the dir that has the .env files


wanderer:
I think that is the same behavior vite has though (you need to run it in the dir that has the .env and vite config)

wanderer:
closing for now

PR #576, opened 2025-06-02
Use Sentry's setTag, setContext and improve general logging behaviour

### LINKED ISSUE #543: [error handling] Configure Sentry to accept / pass context data when app throws errors
Extracted from https://github.com/masslbs/Tennessine/issues/516 and related to https://github.com/masslbs/Tennessine/pull/540/files

> passing object ID to glitchtip for better filtering, more context on which order, listing, etc we are seeing errors for
> 
> re how to passing context data see:
> 
> * https://docs.sentry.io/platforms/javascript/enriching-events/tags/
> * https://docs.sentry.io/platforms/javascript/enriching-events/scopes/
> 
> So ideally we attach the data like shopId, keycard public key and order IDs, to a session scope

lbsal:
This commit demonstrates how to use a context object attached to a logtape handler, such that the keys of the context object can be used to fluff up error traces before they hit glitchtip using the Sentry API's `setTag`.

See ListingDetail.tsx for how the logger is set up initially, and logtape-sentry-sink/mod.ts for the usage of setTag.

We also set up a list of predefined keys, limiting the invocation of setTag to those keys. These keys be data that is useful for debugging crash reports, such as an orderId.

closes #543

### DIFF
--- a/packages/frontend/src/main.tsx
+++ b/packages/frontend/src/main.tsx
@@ -34,6 +34,19 @@ const sentryConfig: Config<string, string> = {
   ],
 };
 
+const getMassBrowserDebug = () => {
+  const params = new URLSearchParams(globalThis.location.search);
+  return params.has("mass-debug") && params.get("mass-debug") === "true";
+};
+
+if (getMassBrowserDebug()) {
+  sentryConfig.loggers.push({
+    category: ["mass-market", "relay-client"],
+    sinks: ["console"],
+    lowestLevel: "debug",
+  });
+}
+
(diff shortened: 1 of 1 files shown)


cryptix (commented):
Could we maybe expose `enableLogging()` via `globalThis` or a query argument or something?

I'd be fine with hiding debug by default if we have a way to enable debug logging on production builds (sepolia or mainnet), without making a new build. Basically the equivalent of `MAX_LOGS` for the browser

wanderer (approved):

Issue #336, opened 2025-04-14
QA ERC20 support


cryptix:

cryptix:
I tested paying via CREATE2 deposit address / QRCode today, and that did work.

I'm not really sure how to proceed with the error above. I manually tested `allowance` via [blockscout](https://eth-sepolia.blockscout.com/address/0x25465C55c5bE9304C4161aDC7998C1C132308D1C?tab=read_write_contract) and that works, too.

PR #577, opened 2025-06-03
Add test for erc20 checkout

cryptix:
updates #336


cryptix:
Turns out we were using the wrong contract address for allowance as well as setting it for the wrong address.

PR #579, opened 2025-06-04
State manager hook uses tanstack's useQuery. Create BrowserLevelStore

youngbbcoconut:


youngbbcoconut:
I think we will have to create a wrapper around this hook for returning setters and getters for a given path

packages/react-hooks/useStateManager.ts
```diff
@@ -0,0 +1,62 @@
...
+      ? async () => {
+        const defaultState = new Map(Object.entries({
+          Tags: new Map(),
+          Orders: new Map(),
+          Accounts: new Map(),
+          Inventory: new Map(),
+          Listings: new Map(),
+          Manifest: new Map(),
+          SchemeVersion: 1,
+        }));
```
wanderer, line 41: I think default state belongs in the `schema` module

packages/react-hooks/useStateManager.ts
```diff
@@ -0,0 +1,62 @@
...
+        const dbName = `${keycard.privateKey}-${shopId}`;
```
wanderer, line 42: use the publicKey instead of the private

Issue #583, opened 2025-06-11
[multi-tabs] Detect if other shop tabs have been closed and automatically close modal


lbsal:
In #411 we wanted to throw up an error message to warn the user about the situation. As part of follow-on work, we could potentially resolve the situation dynamically by making use of either [BroadcastChannel](https://developer.mozilla.org/en-US/docs/Web/API/BroadcastChannel) or the [Web Locks API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Locks_API)

---

some extra details on how to represent the state for resolving multiple tabs and potential errors states over BroadcastChannel - parts might be superfluous due to discovering Web Locks
<details>
<summary>states and transitions</summary>
v2
```
status = "authoritative" | "unconnected" | "seeking" | "error" | "blocked" | "recoverable_error" | "fatal_error"

initial status -> "unconnected"

authoritative:
    set if the currently focused tab has an ongoing authenticated relay
unconnected:
    false if the connection to the relay has not been made or has been terminated
seeking:
    the active tab is trying to find other active tabs and waiting for a response
error:
    some network error occured
blocked:
    not authoritative and is communicating with another shop tab

state transitions:

unconnected -> seeking

seeking -> authoritative
seeking -> blocked
seeking -> seeking

authoritative -> error
authoritative -> unconnected

blocked -> unconnected

error -> recoverable_error

error -> fatal_error

recoverable_error -> unconnected
```

if unconnected -> `seeking`

* check if broadcast channel exists otherwise instantiate it
* broadcast `type: register`

if seeking & seek timeout expires -> `authoritative`

* if no responses were received, become authoritative and call relayClient.authenticate
* if response was received but none were authoritative:
    * broadcast `type: roll` and include whether the current tab is focused
    * if only one tab message contains "focused: true", that tab becomes authoritative
    * if more than one tab contains "focused: true", then the tab with the highest roll becomes authoritative
    * if all rolls are equal, then the roll with the latest timestamp is chosen to become authoritative

if seeking and receive response -> `authoritative | blocked | seeking`

* if response is from not authoritative tab, stay `seeking`
* if response is from authoritative tab -> become `blocked`

if trying to become authoritative and fail -> `error`

if error determine if error is recoverable or fatal -> `recoverable_error | fatal_error`
</details>

Issue #385, opened 2025-04-24
if the shop is already open in one browser tab, trying to open it in another tab fails


youngbbcoconut:
[high] if the shop is already open in one browser tab, trying to open it in another tab fails without any error message explaining why → console says 'authentication failed with code 5: already connected from this keycard in another session'. if you then close both tabs and try to load the shop again, it works fine.

PR #584, opened 2025-06-11
feat: catch relay auth error and throw up modal blocking interaction

### LINKED ISSUE #411: Display error message if shop opened in more than one tab
Related to #385 

This is the first version of fixing #385. Ideally, the problem is solved entirely. Until we get there, we'll show an error message.

lbsal:
This commit catches the error that originates from `/packages/client` on calling `authencation` when another session is already in process.

Previously this situation would not be signaled to the user, which was problematic as any subsequent interactions (e.g. cart changes) would not be saved due to not having a valid connection with the relay.

closes #411
related #385 #583


wanderer (changes requested):

wanderer:
except for the comment, lgtm!

PR #587, opened 2025-06-12
feat(local-testnet): run shop generator

### LINKED ISSUE #564: populate shop when running vite dev
Ideally `genShop.ts` would be called to generate a shop when the vite dev server starts. The url for the shop should also be displayed

cryptix:
fixes #564 

* shop uses ID 0x1234
* it persists across restarts of testnet


cryptix:
hmmm... somehow introduced some flakyness in the tests.. switching it back to draft for now

PR #594, opened 2025-06-21
include db in MassMarketConfig instead of passing it as a param to useStateManager

youngbbcoconut:


packages/react-hooks/useStateManager.ts
```diff
@@ -29,7 +29,6 @@ const BrowserLevelStore = (dbName: string) =>
  * db.close() is called on beforeunload to save the keycard nonce for any writes.
  */
 export function useStateManager(params?: {
-  db?: AbstractStore;
```
youngbbcoconut, line 32: We were originally passing in db as the param here for testing this hook. But frontend tests that are testing components that use this hook also need a way to specify which db to use for tests. So we need a way to pass in the db from more top level.

packages/frontend/src/testutils/mod.tsx
```diff
@@ -233,7 +234,7 @@ export const createRouterWrapper = async ({
             <MassMarketProvider
               stateManager={stateManager}
               relayClient={relayClient}
-              config={env}
+              config={{ db: new LevelStore(), ...env }}
```
youngbbcoconut, line 237: All tests to use MemoryLevel db.

packages/react-hooks/_createWrapper.tsx
```diff
@@ -87,7 +87,7 @@ export function createWrapper(
 
     return (
       <WagmiProvider config={config}>
-        <MassMarketProvider>
+        <MassMarketProvider config={{ db: new LevelStore() }}>
```
youngbbcoconut, line 90: All tests to use MemoryLevel db.

wanderer (approved):

wanderer:
nice looks go to me!

PR #653, opened 2025-07-25
checkout flow uses react-hooks and tests pass

youngbbcoconut:


packages/frontend/src/components/cart/Cart.tsx
```diff
@@ -59,27 +61,27 @@ export default function Cart({
...
-    if (!currentOrder || !stateManager) return;
-    logger.debug(`Showing cart items for order ID: ${currentOrder.ID}`);
-    stateManager.get(["Orders", currentOrder.ID])
+    if (!activeOrder || !stateManager) return;
+    logger.debug(`Showing cart items for order ID: ${activeOrder.ID}`);
+    stateManager.get(["Orders", activeOrder.ID])
       .then(async (res: CodecValue | undefined) => {
         if (!res) {
-          throw new Error("No order found");
+          return logError("No order found");
```
cryptix, line 69: this is a bit unorthodox. unless logError should actually return something to the next promise, do:

```suggestion
          logError("No order found");
          return;
```

packages/frontend/src/components/cart/Checkout_test.tsx
```diff
@@ -19,10 +19,18 @@ import {
   ShippingRegionsMap,
 } from "@massmarket/schema";
 import { CodecKey, CodecValue } from "@massmarket/utils/codec";
-import { createRouterWrapper, testClient } from "../../testutils/mod.tsx";
+import { testClient } from "../../testutils/mod.tsx";
 import Checkout from "./Checkout.tsx";
 import ShippingDetails from "./ShippingDetails.tsx";
 import ChoosePayment from "./ChoosePayment.tsx";
+import {
+  createTestRelayClient,
```
cryptix, line 27: are we using two test clients now? 🤔

packages/frontend/src/components/cart/Checkout_test.tsx
```diff
@@ -33,370 +41,440 @@ type CurrencyTestConfig = {
...
+  expectedTotalAmount: "0.0035 ETH",
+};
 
-  // Remove merchant's keycard to free up for customer
-  localStorage.removeItem(`keycard${shopId}`);
+const EDD_CONFIG: CurrencyTestConfig = {
+  name: "ERC20 (EDD)",
+  address: abi.eddiesAddress as `0x${string}`,
+  expectedSymbol: "EDD",
+  isContract: true,
```
cryptix, line 56: well.. this _should actually_ be false but i doesn't really matter for the test as-is.

(with `isContract:true`, the payment system assumes the payee is another contract that will process the payment. With `false` the amount is just deposited in the address/account)

cryptix (approved):
LGTM! Just a few nitpicks

packages/frontend/src/components/cart/Checkout_test.tsx
cryptix: ah, i see. `testClient` has the `readContract` helpers

PR #686, opened 2025-08-18
Creates useMyShops query to fetch all merchant's shops and the shop's metadata 

devjinyoung:


packages/react-hooks/useMyShops.ts
```diff
@@ -0,0 +1,75 @@
...
+        while (i < balance) {
+          const token = await tokenOfOwnerByIndex(shopPublicClient!, [
+            address!,
+            i,
+          ]);
+          const data = await getShopData(token);
+          allShops.push({ ...data, id: toHex(token) });
+          i++;
+        }
+        return allShops;
```
wanderer, line 65: This runs sequentially while we could be doing it concuntly . 
```suggestion
        const ShopDataPromises = []
        while (i < balance) {
          const dataPromise = tokenOfOwnerByIndex(shopPublicClient!, [
            address!,
            i,
          ]).then(tokenId => getShopData(tokenId))
          ShopDataPromises.push(ShopDataPromises);
          i++;
        }
        return Promise.all(ShopDataPromises);
```

wanderer (changes requested):

PR #701, opened 2025-08-27
out of stock UI and updated tests

devjinyoung:
In the cart view, if there are items in the cart that  we no longer have enough inventory for, then we display the out of stock message. Then, when user tries to commit the order, we check that all the items in the order have enough stock. If there are any items with not enough stock, we remove the item from the order then proceed with committing the item.  

You can just look at the Cart.tsx file


packages/frontend/src/components/cart/Cart.tsx
```diff
@@ -120,6 +131,21 @@ export default function Cart({
         logger.debug("orderId not found");
         throw new Error("No order found");
       }
+      const items = activeOrder.Items;
```
devjinyoung, line 167: We now check that each item is in stock before committing the order. If there are any items that we do not have enough inventory for, then we remove the item from the order.

packages/frontend/src/components/cart/Cart.tsx
```diff
@@ -94,6 +98,7 @@ export default function Cart({
     const allCartItems: Map<ListingId, Listing> = new Map();
     // Get price and metadata for all the selected items in the order.
     const updatedQtyMap = new Map();
+    const updatedInventoryMap = new Map();
```
wanderer, line 134: please add a subscription to the inventory to keep the InventoryMap update todate

wanderer (changes requested):

